PR #11, opened 2016-01-23
Luke/remove combine reducers

lukewestby:
This removes the need for `combineReducers`. It walks the state tree searching for plain objects and if it finds one that is a `loop`, it will collect the effect and replace the value with just the model. During dispatch a custom action is dispatched to the inner store to replace the raw state containing effects with the parsed state. Subscribed listeners are not triggered during this replacement phase.


cogell:
I like this option so that we can combine `redux-loop` with different `combineReducers`

lukewestby:
The only issue with this is that it only checks plain objects. If you have a custom `combineReducers` implementation that combines things into an `Immutable.Map`, for example, the effects are going to get stuck. For those cases, you'll want to turn this functionality off and compose effects up to the top level of your state manually, like in Elm.

This change is designed for compatibility with tools like https://github.com/raisemarketplace/ground-control, which build on top of Redux and use the built-in `combineReducers` internally.

cogell:
Oh thanks for the heads up! `redux-immutablejs` was exactly the `combineReducers` I was thinking of.

lukewestby:
No problem! I know the documentation on the project leaves much to be desired right now, but I'm planning to include a list of all the known scenarios where effects will get stuck and how to write reducers that work for those scenarios.

Another option I'm considering is allowing to pass a custom `visitNode` function so that we can let people specify manually how to search their state tree for effects. Let me know if you have any input or suggestions!

lukewestby:
Status update:

Currently mulling over whether to provide the ability to provide a custom effect reclaimer function. This would allow someone to write a function like:

```javascript
import { Map } from 'immutable';
import { mapValues, isPlainObject } from 'lodash';

function reclaim(obj, visitNode) {
  if (Map.isMap(obj)) {
    return obj.map(visitNode);
  } else if (isPlainObject(obj)) {
    return mapValues(obj, visitNode);
  } else {
    return obj;
  }
}
```

This will allow folks to use redux-loop seemlessly with anything else in the redux ecosystem at a fairly low potential cost.

gaearon:
Would going over the state not be slow? This looks like a potential perf issue.

lukewestby:
@gaearon that's exactly why I haven't merged it. I'm not totally sure this can be made to be fast. It works fine if you have a small state or if the interesting parts of your state are something other than a plain object, like an array or an `Immutable.Map` or something, because those get detected as leaves. I'd imagine that many if not most folks use plain objects though.

lukewestby:
I'd really prefer to just recommend people use the Elm architecture completely once we can land `Effects.lift` from @jarvisaoieong's contribution

jarvisaoieong:
-1 for this
As @gaearon said, it will have perf issue.
It add some black magic and make the rawState shape awful.

lukewestby:
Sounds good. The original intention here was to allow for easier integration with our other redux-related project, `ground-control`, which does some automation of state shape. The ability to provide a custom `combineReducers` implementation has since been added there so I think it's probably safe to drop this change and leave `combineReducers` in. Relatedly, I think it would be good to provide the option to lift the effects from a `combineReducers` result using `Effects.lift` once we can get that released.

gaearon:
Can you show how `lift` works in this case?

lukewestby:
It might look like this:

```javascript
const nestedReducer = combineAndLiftReducers({
  prop1: prop1Reducer,
  prop2: prop2Reducer,
}, Actions.nestedAction);
```

such that calling `nestedReducer` would always return a sub-model and a sub-effect which has already been lifted to `nestedAction`.

Then you could define a top level reducer like:

```javascript
const appReducer = createReducer({
  // ...
  [Actions.nestedAction]: nestedReducer,
});
```

Now that I'm looking at it actually written down though it's become clear that a nested reducer would require knowledge of a parent action, breaking down the fractal analogy. There's definitely some boilerplate to be eliminated with nesting reducers via the Elm architecture, but I don't think this is the way to go.

This conversation has been super helpful for vetting my bad ideas 😁

lukewestby:
Just had a solid internal discussion about this, definitely closing. Thanks folks!

PR #12, opened 2016-01-27
Use IE8-compatible ['catch'] syntax over .catch()

cormacrelf:
Old versions of IE can't handle properties named catch that are accessed with dot syntax.

### DIFF
--- a/modules/install.js
+++ b/modules/install.js
@@ -52,7 +52,7 @@ export function install() {
           const materializedActions = [].concat(actions).filter(a => a);
           return Promise.all(materializedActions.map(dispatch));
         })
-        .catch((error) => {
+        ['catch']((error) => {
           console.error(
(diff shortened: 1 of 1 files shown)


cormacrelf:
Don't have to merge, because when one is transpiling ES6 or ES7 this is one's own concern.

lukewestby:
Thanks for catching this! We should definitely fix this, but I'd rather have babel handle this during build than having `['catch']` in the source. Do you want to submit another PR to add those babel transforms to the package? You can test output with `npm run build` and check the `lib` directory.

If not, I can do this and publish a new release later today.

lukewestby:
Published 1.0.3

Thanks @cormacrelf!

cormacrelf:
Nice. Strictly, we're only required to support down to ES5, but this is always welcome in promise libraries!

PR #14, opened 2016-02-04
Use modern API and conventions in README

gaearon:


lukewestby:
Thanks so much for doing this for us! I'll try to stay more on top of API changes in docs for future changes.

gaearon:
Hopefully we won’t have more :D

PR #19, opened 2016-02-16
add Effects.lift

jarvisaoieong:
Your latest commit that make all internal effect promises resolve an array is cool. It make the Effects.lift very easy to implement.


lukewestby:
Excellent, thank you!

PR #25, opened 2016-03-01
[WIP] Effects queue with simply postponed listeners

bardt:
Work in progress on https://github.com/raisemarketplace/redux-loop/issues/23. 

Currently no elegant way to postpone listeners to be run after effects management found. Looks like the only way is to completely replace redux store subscription method.

@lukewestby`s review needed.


lukewestby:
So I think I'm more or less following along from a high level - effects are inserted into an array and then removed once they've been processed. The only thing I'm not really understanding is why we need to change when listeners are called. Could you elaborate on that aspect?

bardt:
@lukewestby sure! In the [dispatch function](https://github.com/raisemarketplace/redux-loop/pull/25/files#diff-cb03f5868274bee764df3e04d233383dL44) we first dispatch on vanilla redux store. That means, all listeners will be run synchronously _before_ we process effects and update the queue. So, if we've had no effects in the queue before and get some effects from store after dispatch, and we check queue emptiness in one of store listeners, we get wrong positive, because we have not added scheduled effects to the queue yet.

I hope the last sentence is not too long to understand and I did not mess the grammar. Main idea is to postpone listeners until we update the queue after each dispatch.

bardt:
@lukewestby I've just added an example of server usage as I expect it. It required some changes to the original example, please review. 

Now working on test cases.

BTW, what about my explanation of listeners change? Is it clearer now?

bardt:
@lukewestby OK, I've rebased this branch on current master and changed implementation significantly. Now I don't overwrite original redux subscribtion. Instead, I assume _every_ effect (except for `none`) resolves into actions array and mark the whole effect as finished right before dispatching its last action. Is my assumption right?

lukewestby:
It's possible that someone might dispatch a tree of effects consisting of nothing but empty batches, but i don't think it's important to worry about that. This looks good to me, I'm going to play around with the example and look to merge some time this week. Thanks!

lukewestby:
I'm not sure how this is going to fit in to the next API and I'd rather wait for a version after that to figure it out so I'm going to close this

PR #34, opened 2016-03-20
Pass the received enhancers to the next enhancer

### LINKED ISSUE #28: Route action as effect (react-router-redux)
Hi, I'm struggling with making effects work with react-router-redux. I've setup react-router-redux routerMiddleware and trying to change history as effect.

Example:

``` javascript
import { push } from 'react-router-redux';

function reducer(state, action) {
  switch(action.type) {
  case 'FIRST_ACTION':
    return loop(
      state.set('firstRun', true),
      Effects.constant(push('/delivery'))
    );
}
```

In redux DevTools I see that action @@router/CALL_HISTORY_METHOD is triggered, but history (url) is not changed. Trying to figure out, what I'm doing wrong here.

claudiuandrei:
When not the outer enhancer, the next enhancer loses information about the previous ones. In order to fix this issue, we just need to pass the enhancer parameter if received.

### DIFF
--- a/modules/install.js
+++ b/modules/install.js
@@ -31,9 +31,9 @@ function liftReducer(reducer) {
  * attached to the current model as established by the original dispatch.
  */
 export function install() {
-  return (next) => (reducer, initialState) => {
+  return (next) => (reducer, initialState, enhancer) => {
     const liftedInitialState = liftState(initialState);
-    const store = next(liftReducer(reducer), liftedInitialState);
+    const store = next(liftReducer(reducer), liftedInitialState, enhancer);
 
     function dispatch(action) {
       const dispatchedAction = store.dispatch(action);


lukewestby:
Excellent! Thanks for fixing this, it's been bugging me.

claudiuandrei:
@lukewestby can you hold on on merging this, I think it does not solve the problem, sorry for the false hope, but I'm still looking into it.

claudiuandrei:
@lukewestby this is not solving #28 but it actually allows the enhancer to be passed if somebody wants to use that parameter. Currently redux enhancers are passing it.

lukewestby:
Yep, good idea. We'll merge it in even though it doesn't help #28. Let me know if you can find out any way around that issue besides recommending putting `install()` first in the chain.

claudiuandrei:
As far as I understand how the enhancers are working, there is no real way to dispatch from the top, unless the top is owned by redux-loop.

PR #35, opened 2016-03-20
Remove the redux-saga examples, they are confusing

claudiuandrei:
I think these examples were added by mistake, they are confusing as they are not redux-loop examples.


lukewestby:
oh shoot, not sure how those got in there. thanks!

PR #36, opened 2016-03-20
Remove mapValues from combineReducers

claudiuandrei:
mapValues should not have side-effects, so the code can either be imperative or it should be using a reduce that always returns a loop. This is using an imperative way as redux function does the same thing.


lukewestby:
Looks good! Always happy to receive perf-improving PRs :smile:

leonaves:
Sorry if something is lost on me but how did mapValues previously have side effects? It never modified the value of reducerMap, it reduced down based on a new initial object (`{}`).

PR #37, opened 2016-03-20
Add call effect type

### LINKED ISSUE #30: Suggestion: Add Effects.call type
Some sync tasks such as access localStorage, call [history](https://github.com/mjackson/history) method are also side effect.

Now we only can use the Effects.promise to do some stuff. It works but a bit strange.

e.g.

```js
// reducer
export default (state, action) => {
  if (action.type === 'SAVE_ACCESS_TOKEN') {
    return loop(
      state
    ,
      Effects.promise(saveAccessToken, action.accessToken)
      // Effects.call(saveAccessToken, action.accessToken)
    );
  }
  return loop(state, Effects.none());
}
```

```js
// tasks
export saveAccessToken = (accessToken) => {
  localStorage.accessToken = accessToken;
  return Promise.resolve({type: 'SAVE_ACCESS_TOKEN_SUCCESS'});
  // return {type: 'SAVE_ACCESS_TOKEN_SUCCESS'};
}
```

claudiuandrei:
Allow to pass sync functions that are doings something and then trigger an action. Fixes #30.

### DIFF
--- a/modules/effects.js
+++ b/modules/effects.js
@@ -79,8 +79,8 @@ export function promise(factory, ...args) {
 
 /**
  * Creates an effect for a function that returns an action.
- * @param {Function} factory The function to invoke with the given args an action.
- * @returns {Object} The wrapped effect of type PROMISE.
+ * @param {Function} factory The function to invoke with the given args that returns an action.
+ * @returns {Object} The wrapped effect of type CALL.
  */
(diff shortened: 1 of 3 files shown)


lukewestby:
@claudiuandrei can you add a test case for this?

lukewestby:
Looks good, thanks!

Issue #38, opened 2016-03-24
combineReducers does not support root immutable state


leonaves:
The implementation of combineReducers provided with this library does not support a root immutable state such as Map. (Which is odd as the example in the repo uses a Map as its root state—but doesn't use `combineReducers`). The problematic areas are here in `combineReducers.js`:

```javascript
// line 22
const previousStateForKey = state[key];
```

And the mapValues function in `utils.js`:

```javascript
export function mapValues(object, mapper) {
  return Object.keys(object).reduce((current, key) => {
    current[key] = mapper(object[key], key);
    return current;
  }, {});
}
```

This limits the root state to being a plain object. I have written my own version of these two functions for my project, but it's a bit awkward as I have to use isLoop from the library like so:

```javascript
import { isLoop } from '../../../node_modules/redux-loop/lib/loop';
```

Which obviously could break at any time as this isn't an exposed function. Ideally, support for this would be built in. I was thinking about opening a pull request, but I wasn't sure how you might want to go about this (if you wanted it at all). Obviously I could drop in my function and export it as "`combineReducersImmutable`" or something similar. Or I could adjust the original function to check for immutable root states and to act appropriately.

Both these add code for one specific library however, which I've never been a huge fan of (but perhaps I'm overthinking it—it is a fairly popular library). So the last thing I thought of was to adjust combineReducers so it can accept any root data structure, something like:

```javascript
export function combineReducers(reducerMap, initialState = {}, accessor, mutator) {
    // code...
}

// and then to use:
const appReducer = combineReducers(reducers, Map(), 'get', 'set');
```

I may be jumping the shark here however (the accessor and mutator function names as strings makes me cringe—perhaps there's a better way? I'm just spitballing with this last suggestion really).

Anyway, would be interested to hear your thoughts, and happy to submit a pull request for an approach you prefer.

leonaves:
Just thinking a bit more, another option might be to either export Symbol(isEffect) or to use Symbol.for() instead, as this is the main reason the function can't be replicated without requiring isLoop. (isLoop would be trivial to recreate if it didn't need that instance of that symbol).

lukewestby:
I don't want to export `Symbol(isEffect)`, it's 100% an implementation detail. Exposing `isLoop` is definitely an option, I can think of plenty of other reasons why someone might want to have this function available in addition to this one.

[...]

PR #40, opened 2016-03-25
Addresses #38, combineReducers can handle custom data structures.

leonaves:
I returned to a reduce function on the object.keys, as I don't see it being any less efficient, and personally I much prefer it to a for loop, but I kept some of the changes from #36 which did increase efficiency. Let me know what you think.

### DIFF
--- a/modules/combineReducers.js
+++ b/modules/combineReducers.js
@@ -12,32 +12,36 @@ function optimizeBatch(effects) {
   }
 }
 
-export function combineReducers(reducerMap) {
-  return function finalReducer(state = {}, action) {
-    let hasChanged = false;
-    let effects = [];
-    let model = {};
-
-    const reducerMapKeys = Object.keys(reducerMap);
-    for (let i = 0; i < reducerMapKeys.length; i++) {
-      const key = reducerMapKeys[i];
-      const reducer = reducerMap[key];
-
-      const previousStateForKey = state[key];
-      let nextStateForKey = reducer(previousStateForKey, action);
-
-      if (isLoop(nextStateForKey)) {
-        effects.push(getEffect(nextStateForKey));
-        nextStateForKey = getModel(nextStateForKey);
-      }
-
-      model[key] = nextStateForKey;
-      hasChanged = hasChanged || nextStateForKey !== previousStateForKey;
+function combineReducers(reducerMap, rootState = {}, accessor, mutator) {
+    if (typeof accessor !== 'function') {
+        accessor = (child, key) => child[key];
     }
+    if (typeof mutator !== 'function') {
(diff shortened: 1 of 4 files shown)


modules/combineReducers.js
```diff
@@ -12,32 +12,36 @@ function optimizeBatch(effects) {
   }
 }
 
-export function combineReducers(reducerMap) {
-  return function finalReducer(state = {}, action) {
-    let hasChanged = false;
-    let effects = [];
-    let model = {};
+export function combineReducers(reducerMap, rootState = {}, accessor, mutator) {
+    if (typeof accessor !== 'function') {
```
lukewestby: can you lift the default `accessor` and `mutator` functions out of the `combineReducers` body and use default args to assign them? could be a little bit of a perf boost to not create new default functions on every call

leonaves:
How's that, I'm not too familiar with code style for long argument lists in javascript so I just followed [PSR-2](http://www.php-fig.org/psr/psr-2/#4-4-method-arguments) 😜 I won't be offended if you reformat it.

leonaves:
How's that? I'm not too familiar with code style for long argument lists in javascript so I just followed [PSR-2](http://www.php-fig.org/psr/psr-2/#4-4-method-arguments) 😜 I won't be offended if you reformat it.

lukewestby:
Code style looks good! Can you install `immutable` as a devDependency and write a test case for the new functionality to show that (a) the default implementation works the same and (b) your new use case with Immutable works like you expect? You can put those tests in a separate file `combineReducers.test.js`

As a side note, does `Object.keys(obj)` work on an `Immutable.Map()`?

leonaves:
Okay, getting on that now, and now `Object.keys(obj)` does not work on `Immutable.Map()`, no, it will give the keys of the root object, e.g.:

```javascript
[ 'size',
  '_root',
  '__ownerID',
  '__hash',
  '__altered' ]
```

But that's okay, `reducerMap` should be an object anyway, it's just the state itself that may be another data structure.

leonaves:
You could change the `Object.keys(reducerMap).reduce()` to a [`for...of` loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of), which would support a reducerMap of any data structure that implements the Iterable interface (including `Immutable.Map()`) but I'm not sure how necessary that is.

lukewestby:
Aha! I'm following now, I think we'll stick with the plain object reducer map. Thanks for clearing that up. This looks good to me! I'm planning to cut a release tomorrow, I'll make sure this gets merged in for that.

lukewestby:
Awesome! @leonaves this is an exemplary OSS contribution, thanks so much.

PR #43, opened 2016-04-04
Fix header links in ApiDocs.md

leonaves:


lukewestby:
ugh i always get those wrong. thanks!

PR #63, opened 2016-06-15
export isLoop and none for easier changes to combineReducers

emmby:
based on https://github.com/raisemarketplace/redux-loop/pull/53 but also exports none, which is needed by combineReducers.optimizeBatch

### DIFF
--- a/modules/index.js
+++ b/modules/index.js
@@ -3,6 +3,7 @@ import {
   liftState,
   getModel,
   getEffect,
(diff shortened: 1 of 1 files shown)


bdwain:
none is already exported as part of Effects

emmby:
True fact.

lukewestby:
@emmby thanks for your contribution anyway!

PR #65, opened 2016-06-16
FIX #64: Fix dropping of effects by detecting if an effect is already scheduled.

### LINKED ISSUE #64: Effects not being executed when multiple dispatches in the same tick
I've been working with redux loop and have been loving it, although I think I've found a bug where if any dispatch occurs in the same tick of the event loop after a dispatch which creates an effect, the effect will not be executed.

Specifically I've experienced this with redux-form, which will dispatch actions on React lifecycle hooks (i.e. componentDidMount and componentWillRecieveProps).

I'm working on creating a minimal reproducible scenario, but comments are welcome.

I believe it has something to do with https://github.com/raisemarketplace/redux-loop/blob/master/modules/install.js#L31, although I'm still digging.

akiellor:
This is a potential fix to the issue #64.

There is a test that simulates the behaviour specified in #64 where an action causes an effect, and a store subscriber (e.g. react redux) ends up triggering a component hook which dispatches another action, resulting in the original effect being clobbered with a `NONE` effect.

Please let me know of any improvements I can make to this PR!

### DIFF
--- a/modules/effects.js
+++ b/modules/effects.js
@@ -53,6 +53,15 @@ export function isEffect(object) {
 }
 
 /**
+ * Determines id the effect object is of type none
+ * @param {Object} The object to inspect.
+ * @returns {Boolean} Whether the object is a none effect.
+ */
+export function isNone(object) {
+  return object ? object.type === effectTypes.NONE : false;
+}
+
+/**
  * Creates a noop effect.
  * @returns {Object} An effect of type NONE, essentially a no-op.
  */
--- a/modules/install.js
+++ b/modules/install.js
@@ -13,6 +13,7 @@ import {
   batch,
   none,
   isEffect,
+  isNone,
   effectToPromise,
 } from './effects';
 
@@ -28,7 +29,11 @@ export function install() {
     const liftReducer = (reducer) => (state, action) => {
       const result = reducer(state, action);
       const [model, effect] = liftState(result);
-      currentEffect = effect;
+      if (isNone(currentEffect)) {
+        currentEffect = effect
+      } else {
+        currentEffect = batch([currentEffect, effect]);
+      }
       return model;
     };
 
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -97,6 +97,40 @@ test('a looped action gets dispatched after the action that initiated it is redu
     });
 });
 
+test('Effects are proccessed in the order they are recieved', (t) => {
+  t.plan(1);
+
+  const reducer = function (state = {}, action) {
+    switch (action.type) {
+      case 'LOOPED_BEGIN':
+        return loop(
+          {...state, done: false},
+          Effects.constant({type: 'LOOPED_END'})
+        )
+      case 'LOOPED_END':
+        return {...state, done: true};
+      default:
+        return state
+    }
+  };
+
+  const store = finalCreateStore(reducer, {});
+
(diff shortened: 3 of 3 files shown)


akiellor:
@lukewestby Any feedback for this PR. I'm happy to make any changes.

lukewestby:
this looks good. i'll merge it in this week with some other fixes. it's possible i might make some minor modifications in order to make sure everything is coherent but overall this is a great catch and a great fix. thanks so much!

akiellor:
Awesome! Thanks.

bdwain:
could it cause problems to arbitrarily batch events together? if you have a relatively quick effect and then you batch a long running effect with it, the quick effect won't dispatch its result until the long one is done. 

maybe it'd make more sense to implement some kind of queue?

bdwain:
also, i'm noticing effects getting dropped if the dispatch that causes the effect also creates a new state that causes an error in my react component. i'm not sure if that's related to this or not (or even something we care about). but the changes here did not fix that issue.

akiellor:
@bdwain re:batching effects, at the moment we drop effects on the floor, batching is better. The only semantic I think is important is that the promise returned for the initial action should not resolve until all the effects created (either directly or by triggered actions) also resolve. This could be modeled as a batch or a sequence.

@lukewestby Is this PR likely to become not required given the new API change work going on? If the change is desirable, it would be nice to get a patch release out just to get off our own fork.

lukewestby:
I can imagine situations where folks would experience performance issues if a long-running effect gets paired with a quick one. @akiellor, would you mind implementing this change with a queue as @bdwain suggested? (FWIW redux-loop 3.0 will use a queue to deal with this as well). Once that's in place I'll be happy to do a patch release so y'all can get off of your fork.

Imater:
This Pull request fixes error multiply start dispatch in one tick. 
Please recieve the pull request. We realy need it.

lukewestby:
@Imater it will be merged and released as a patch once it's been implemented as described

akiellor:
I'll try explain the implementation in this branch with the following scenario:

Let A, B and C be actions all of which cause effects denoted as A', B', C'

A and C are dispatched directly through normal store usage.
A dispatches B via a store subscription hook.

Current behaviour as of 2.2.1:

* Effect A' is not executed (bug)
* B' and C' are both executed
* Promise for A resolves when B' completes
* Promise for C is independent and resolves separately

Implemented behaviour:

* All effects are executed
* Promise for A resolves when A' and B' complete (A' and B' are batched)
* Promise for C is independent and resolves separately, C' is NOT batched with A' or B'

----

The reason I raise this is that I'm unsure what a queuing implementation would look like. And I'm not sure what the implications would be to the Promise returned for Action A.

My expectation is as it is implemented, if an action causes another action then they are batched. Otherwise they are not.

I'd be happy to add a test which demonstrates this behaviour more clearly if it helps.

lukewestby:
I spent some time thinking about this last night and it seems like it would be more beneficial to merge it and release it as is. Since the queueing mechanism will be available in the next version in a few weeks I think it makes more sense for folks to have this working now and potentially have to work around the rare performance problem with a `setTimeout` than it does to make folks continue to work around the real underlying issue all the time.

lukewestby:
Released as 2.2.2

akiellor:
Thanks @lukewestby

Issue #66, opened 2016-06-21
New Effect API to allow for reusable side-effect methods


bdwain:
Copied this comment from #54 

After using redux-loop a bit more in practice, I've started to have some thoughts on how the api could encourage a little more code reuse and simplify a few things.

One issue with forcing effects to return/resolve with redux actions is that the implementations of the side-effects become tied to redux/redux-loop. Suddenly a pretty standard function to make a request, such as this,

```js
import apiClient from './apiClient.js'; //to abstract away fetch details like json parsing and error codes

function fetchItems(userid){
  return apiClient(`/users/${userId}/items`)
      .then(rsp => {
           //process rsp
           return processedRsp;
       }).catch(err => {
            console.log(':(');
            return err;
       });
}
```

can't be used by redux-loop without changing it in a way that makes it unusable by something else (you can no longer reject the promise, and your resolution needs to be a redux action and not just the list of items).

Additionally, even if you wanted it only for redux-loop, it would be impossible (or at least awkward) to use a side-effect to return different types of actions depending on how it is called (I haven't had a need to do this yet, but it's not hard to imagine a case where you'd want it)

What if we allowed side-effects to become completely agnostic to redux-loop/redux in general, and instead return whatever they felt was necessary to complete their purpose (so for async methods, prob would be a promise that resolves or rejects to indicate success).

So instead of defining the actions triggered (or not) by the effect in the method, it could be defined in the Effect object itself.

The new effect objects returned by a reducer might look like this

```js
//actions.js
function itemsLoadedActionCreator(items){
    return {
       type: 'ITEMS_LOADED',
       items: items
    };
}
// ... other action creators 

//reducer.js
return loop(newState, Effect.promise(fetchItems, itemsLoadedActionCreator, failHandler, params...));

Effect.call(writeToLocalStore, syncMethodActionCreator, params...)

//Effects.constant could be implemented as syntactic sugar for 
Effects.call(null, constantAction());

//batch would be basically the same
//I imagine lift would be similar, though I haven't looked into it much yet
```

And now your effects are completely reusable and agnostic to your choice of state management. You could swap out redux-loop or even redux iteself without having to change them.

bdwain:
Additionally, and it could be separate from this, I think it could be useful to allow passing the dispatch method to side-effects. I have a few scenarios where it would be nice if the side effect was able to dispatch an action, without it being the specific action that is the result of the effect.

lukewestby:
[...]

PR #69, opened 2016-06-23
initial work on new api

lukewestby:
as per #66


modules/cmd.js
```diff
@@ -0,0 +1,150 @@
...
+        .then((action) => [action])
+
+    case cmdTypes.CONSTANT:
+      return Promise.resolve(cmd.action)
+
+    case cmdTypes.BATCH:
+      const filteredCmds = cmd.cmds
+        .filter((nextCmd) => nextCmd.type !== cmdTypes.NONE)
+
+      const batchedPromises = filteredCmds.map(cmdToPromise).filter((x) => x)
```
bdwain: i don't think you need to filter above and then filter here also. seems like the only promise that can be falsey is none. you can probably just say `batchedPromises = cmd.cmds.map(cmdToPromise),filter(x => x)`

lukewestby: yeah i'll get rid of the first filter. nested empty batches will also return `null` so both cases need to be handled.

bdwain: i think they'll be handled by the line below `if (batchedPromises.length === 0) return null` right?

modules/loop.js
```diff
@@ -1,76 +1,34 @@
...
-      isEffect(effect),
-      'Given effect is not an effect instance.'
-    );
+      isCmd(cmd),
+      'Given cmd is not an Cmd instance.'
+    )
   }
 
-  return [model, effect];
+  return [model, effect]
```
bdwain: this should be [model, cmd] i beleive

modules/install.js
```diff
@@ -1,67 +1,68 @@
...
+      cmdsQueue.push({ originalAction: action, cmd })
+      return model
+    }
+
+    const store = next(liftReducer(reducer), initialModel, enhancer)
+
+    const runCmds = (queue) => {
+      const promises = queue.map(runCmd).filter((x) => x)
+      if (promises.length === 0) return noCmdPromise
+      else if (promises.length === 1) return promises[0].then(() => {})
```
bdwain: why the empty promise handlers?

modules/combineReducers.js
```diff
@@ -1,55 +1,45 @@
...
-            optimizeBatch(effects)
-        );
-    };
+export const combineReducers = (
+  reducerMap,
+  rootState = {},
+  accessor = defaultAccessor,
+  mutator = defaultMutator
+) => {
+  return const finalReducer = (state = rootState, action) => {
```
bdwain: error on this line

bdwain:
this looks like all of the cmds have optional action creators. did you mean to do that or do you still want a separate thing where they're optional and have cmds require them?

modules/install.js
lukewestby: i'm not sure what the promise should actually resolve at this point so i've left it blank for now

bdwain: got it

lukewestby:
right now it's in a state where we could go either way. i still like the idea of being explicit with arguments but i think something like `Cmd.promise(createPromise, undefined, undefined, ...args)` could be okay as well

PR #79, opened 2016-07-18
3.0

lukewestby:
This is the branch for all the changes that will be released as redux-loop 3.0. Feel free to start discussions and ask questions here, although I may ask that discussions be continued in separate issues that we can reference in the description here.

TODOs (will be marked as done once all discussion is complete _and_ code changes are in)
- [x] New effects API - `Cmd`
- [ ] Migrate to TypeScript
- [ ] Effects that do not have to return actions - `arbitrary` or `port` with async versions for proper ordering
- [ ] `batchedSerial` or `sequence` effect
- [ ] Documentation changes for making a recommendation about server-side rendering


bholloway:
Nice work on `redux-loop` and good to see there is ongoing development.

I intend to leverage the [elm-like reducer signature](https://twitter.com/dan_abramov/status/630894436810158080) in another library and will be recommending `redux-loop` as the implementation to use.

Can you please tell me whether `3.0` will contain changes to how reducers are composed?

Right now I can achieve everything I need with `combineReducers()` and `getModel()`. To use my library the user will pass in these methods. Is there any chance that these methods will no longer available?

lukewestby:

PR #80, opened 2016-08-03
Update README (elm-effects is no more)

lynn:
`elm-effects` got moved into `core`, and now commands are part of the language itself. Neat!


lukewestby:
Excellent point! 😄  There is a current effort to migrate redux-loop to be more similar to the Elm `Cmd` API, so this ill be merged once that effort is complete. As it stands now, this API still more closely resembles the 0.16 system.

Issue #56, opened 2016-05-19
add batchSerial effect type?


bdwain:
I've run into a few scenarios where i want to perform multiple effects as a result of a state change, but have then run in order instead of in parallel. Right now, the best option I see to do that is using Effect.call like this

```javascript
function reducer(...){
  //state changes...
  const effectFn = () => {
    action1();
    action2();
  };

  return loop(newState, Effects.call(effectFn));
}
```

but you can't reference that function in a test. And while you could export effectFn somewhere so it could be referenced in a test, it could get out of hand quickly if you had to start exporting lots of difference combos of actions.

I thought it could be nice to have a batchSerial (or just serial?) action that works the same as batch, but runs everything in order and waits for the previous effect to finish

```javascript
return loop(newState, Effects.batchSerial([
   Effects.call(action1],
   Effects.call(action2)
);
```

lukewestby:
This is a good idea, I think we should add it. Do you want to submit a PR? Let's not worry about the name for now, we can bikeshed once we've got something that works 😃

lukewestby:
Actually, I think `Effects.sequence(effects: Array<Effect>)` would be good as it matches nicely against Elm's `Task.sequence`. What do you think?

bdwain:
i think sequence makes sense.  i'll go with that.

i've had [this implementation](https://github.com/bdwain/redux-loop/commit/692225c54382633ab33ae983a218464ba356c933) in a branch in my fork for a while. (note there's a typo on line 40 in that commit. batch should be serial)

what do you think? i don't see a simple way to turn it into a one liner like the others, since i feel like it needs this recursive style of pulling an effect off the list and then batching the rest, since each promise resolution needs to trigger what is essentially a batch of the remaining effects. this seems to work for me though.

bdwain:
i'll get something working for this in a branch of 3.0

lukewestby:
@bdwain awesome, thank you! i'll have proper merge permissions set up for you by tomorrow at the latest

bdwain:
awesome. thanks!

avocade:
+1

hannupekka:
Any idea when this will be implemented?

PR #81, opened 2016-08-04
add sequence cmd

bdwain:
see #56


modules/install.js
```diff
@@ -1,6 +1,7 @@
 import { throwInvariant } from './utils'
 import { loop, getCmd, getModel, isLoop, liftState } from './loop'
 import Cmd, { cmdToPromise, isCmd } from './cmd'
+import { loopPromiseCaughtError } from './errors'
```
bdwain: just noticed this was causing an error

bdwain:
thoughts @lukewestby ?

lukewestby:
The next version is going to be based on a `Task` type, it'll be no problem to define a `Task.sequence` function on that type

Issue #67, opened 2016-06-21
Allow for passing dispatch to side-effect methods


bdwain:
It'd be nice if the side-effects could have access to the dispatch method to dispatch arbitrary actions that are not tied to the specific task/cmd/effect they are running at the time. 

Here's a scenario where this would definitely be helpful. 

I have an API client in my project that all API requests go through. Its main job is to add auth headers to the requests, but the other thing it does is force log you out when you get a 401 back (because your token expired). With redux-loop right now I typically have to use it like this to check for the 401.

```js
function fetchItems(userid){
  return new Promise(resolve, => {
     apiClient(`/users/${userId}/items`)
       .then(rsp => {
            //process rsp
            resolve(getSuccessAction(processedRsp));
        }).catch(err => {
             resolve(err.status === 401 ? get401ErrorAction() : getItemErrorAction(err));
        });
  });
}
```

The problem is that I have to repeat this 401 logic for every request. And also there's only 1 resolved action, so you can't dispatch the item error if you need the 401 error. 

Ideally, the api client could take a dispatch method, and then just choose to dispatch a 401 error in its internal catch block before the caller handles the promise. And then the caller can also dispatch an error fetching action and not worry about if the 401 happened or not. 

That's just one scenario, but I could see it being useful pretty often.

bdwain:
Also this might have conflicts with #66 so i'm happy to reevaluate it after that is done. I think that is a higher priority

lukewestby:
I think the changes coming with the new `Cmd` API will make this unnecessary. You'd be able to extract that status code logic into a higher-order function and reuse it easily.

```js
const withStatusCode = (actionFor401, actionForOtherErrors) => (error) => {
  return error.status === 401 ? actionFor401() : actionForOtherErrors(error);
};

const actualErrorActionCreator = withStatusCode(actual401ActionCreator, actualOtherErrorCreator);

// ...

Cmd.promise(makeRequest, successAction, actualErrorActionCreator)
```

bdwain:
This removes some of the repetition, but it still requires every Cmd that triggers an api request to know about the 401 functionality, instead of centralizing it in one spot. It could easily be forgotten in a scenario like that, which is my worry.

My thought was something along the lines of providing a getDispatch method (or maybe just the dispatch method directly) with something like this.

```js
Cmd.promise(makeRequest, Cmd.dispatch, successActionCreator, ...etc);

//or maybe
[...]

PR #82, opened 2016-08-04
allow dispatch and getState from cmds

bdwain:
initial implementation idea for #67

### DIFF
--- a/modules/cmd.js
+++ b/modules/cmd.js
@@ -2,6 +2,8 @@ import { throwInvariant, flatten, promisify, isPromiseLike } from './utils';
 
 
 const isCmdSymbol = Symbol('isCmd');
+const dispatchSymbol = Symbol('dispatch');
+const getStateSymbol = Symbol('getState');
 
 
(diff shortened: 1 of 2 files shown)


bdwain:
usage: 

```js
function log1(dispatch){
  console.log('log1');
  dispatch({type: 'log2'});
}

function log2(getState){
  const foo = getState().bar;
  console.log(foo);
}

function handleSomeAction(state){
  return loop(state, Cmd.call(log1, () => {return {type: 'log 1 done'}}, Cmd.getDispatch));
}
```

bdwain:
i've been using this in prod for a bit now. it works really well. it really simplifies a lot of my effects and lets them become more generic side effect functions.

bdwain:
any thoughts on this @lukewestby ?

lukewestby:
I just finished a big rewrite at https://github.com/redux-loop/redux-loop/pull/95

In the notes there's a section on why I don't want to bring in stuff from the rest of the redux API into the redux-loop APIs

PR #89, opened 2016-09-23
Add support for nested loops

evinism:
It would be helpful for `loop` to be agnostic of whether it's passed a loop or a model, to allow composable looped state transitions. This seems to be in line with the style of the rest of the API.  

```
const DoTheThing = (state) => loop(state, Effect1);
const DoTheOtherThing = (state) => loop(state, Effect2);

// such that the following useless reducer would be legal and batch the effects
const reducer = (state, action) => DoTheThing(DoTheOtherThing(state))
```

### DIFF
--- a/modules/loop.js
+++ b/modules/loop.js
@@ -1,5 +1,5 @@
 import { throwInvariant } from './utils';
-import { isEffect, none } from './effects';
+import { isEffect, none, batch } from './effects';
 
 
 /**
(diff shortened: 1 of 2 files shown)


lukewestby:
I'm deprecating a separate `loop` function in favor of just using two-element arrays. Given that, including this would amount to including some helper functions for dealing with arrays and I'd prefer not to have those kind of extras in this library. If you want to release something as a separate library that helps with dealing with arrays for redux-loop I'd be happy to take a look and potentially recommend it in the docs

evinism:
Super cool, I honestly like this move a lot better. I was sort of tricked into thinking that due to the existence of all these super small helper methods, the loop data model was gonna get much more complicated. Glad to see it's moving the other direction.

PR #90, opened 2016-09-23
enhance combineReducers()

bholloway:
There are a couple of features that I need from `combineReducers()`.
* support reducers with `Symbol` keys
* pass additional arguments to reducers

If these enhancements are generally a good thing then please consider my pull request.

I do see that there is a branch to remove `combineReducers()` so this PR may be uninformed. If so please let me know what the near-future plans are and let me know if I can help.

### DIFF
--- a/modules/combineReducers.js
+++ b/modules/combineReducers.js
@@ -23,20 +23,27 @@ const defaultMutator = (state, key, value) => {
   };
 };
 
+const keys = (value) => {
+  return [
+    ...Object.getOwnPropertyNames(value),
+    ...Object.getOwnPropertySymbols(value)
+  ];
+};
+
 export function combineReducers(
     reducerMap,
     rootState = {},
     accessor = defaultAccessor,
     mutator = defaultMutator
 ) {
-    return function finalReducer(state = rootState, action) {
+    return function finalReducer(state = rootState, action, ...rest) {
         let hasChanged = false;
         let effects = [];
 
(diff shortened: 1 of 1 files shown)


bholloway:
**Alternative**

I just took a look at #79 and the WIP on the version `3.0` branch.

The new Effects API is seriously good. `Cmd.batch()` will make it a lot easier to write your own `combineReducers()`.

As an alternative to the PR could I suggest:
* The remaining `3.0` features be pushed to `4.0` so that the new Effects API can be released ASAP.
* Merge the `/remove_combine_reducers` branch into `3.0` if you still want to be rid of it.

**Additional Bike shedding**

If I look at [`combineReducers@3.0`](https://github.com/redux-loop/redux-loop/blob/3.0/modules/combineReducers.js#L33) I see the code:
```javascript
if (isLoop(nextStateForKey)) {
  cmds.push(getCmd(nextStateForKey))
  nextStateForKey = getModel(nextStateForKey)
}
```

If this code can be made more terse then the remainder is next to nothing. Perhaps:
```javascript
nextStateForKey = getModelPushCmd(cmds, nextStateForKey);
```

Just a thought.

lukewestby:
I'm going to close this since I'm considering removing `combineReducers` completely for the next version, but if we decide to keep it around I'll come back, reopen, and figure out a way to get this merged in. Passing additional arguments to reducers is a good strategy for sharing parent state and something I'd like to encourage.

Here's roughly what the next API is going to look like

PR #93, opened 2016-10-06
remove `babel` package

hzoo:
not a big deal but `babel` as a package currently doesn't actually run anything after babel 6 (it's in babel-cli which is a dependency already)

https://github.com/babel/babel/blob/master/packages/babel/cli.js


lukewestby:
@hzoo i'm getting pretty close to a new version so i'll make sure it gets removed during that release

Issue #51, opened 2016-05-14
immutablejs combinereducers example has high overhead


bdwain:
When used with immutablejs as described in the readme, combineReducers looks like it might suffer from high overhead because each time you call the setState method, immutablejs will create a new copy of the store. It would be more efficient if the mutations were batched with the immutablejs method withMutations. (basically the provided combinereducers but with the main block wrapped by "withMutations"

I wouldn't expect a method optimized for immutablejs to be provided by this repo, but as long as it's being mentioned in the readme, it might be worthing pointing out that performance caveat so people know they should probably be writing their own. 

I think something like this would work well

```javascript
function immutableJsCombineReducers(reducerObj) {
  const reducerKeys = Object.keys(reducerObj);

  return (rootState = Map(), action) => {
    let hasChanged = false;
    let effects = [];

    const model = rootState.withMutations(initialState => {
      reducerKeys.reduce((previousState, reducerName) => {
        const currentChildState = previousState.get(reducerName);
        let nextChildState = reducerObj[reducerName](currentChildState, action);

        if (isLoop(nextChildState)) {
          effects.push(getEffect(nextChildState));
          nextChildState = getModel(nextChildState);
        }

        hasChanged = hasChanged || nextChildState !== currentChildState;
        return previousState.set(reducerName, nextChildState);
      }, initialState);
    });

    return loop(
      hasChanged ? model : state,
      optimizeBatch(effects)
    );
  };
}
```

bdwain:
I just realized that the package doesn't export isLoop, so it could be hard to write this. i'll make a PR to export it

bdwain:
closed since #53 was merged

PR #110, opened 2017-03-06
Possible next version API implementation

gyzerok:
This PR contains implementation of the next possible release of the library.

**Some quicknotes:**
 * API have been cut significantly in order to provide simpler experience in learning and using it.
 * It is now written in TypeScript and generates type declarations with itself. This means that types wont get stale.
 * Tests are now done with `mocha`. There is no big decision behind this. It's just more wide spread so more people are already familiar with it.
 * More useful errors for different cases.

**Right now there are following problems:**
 * New API is based on the idea that you will provide single reducer to `createStore`. No more way to use `combineReducers`.
 * You have to provide initial state to `createStore`. This is because of types additions. It's no more sufficient to initiate state using defaults as in Redux itself.

The biggest question right now is how to side-step Redux in a clearer way. I believe that Redux here is used more to piggy back on it's existing infrastructure (DevTools) and we should not follow every other convention of it. In my opinion library should reasonably follow Elm.


lukewestby:
I'm 💯  on no longer implementing `combineReducers`, it is my opinion that it's not really a good idea in the first place. In my Elm apps I keep a fairly flat state shape and I've found that works fine in redux as well. There's definitely more of a "remove boilerplate at all costs" ethos in the JS community but I think it's okay to defy that in a sense.

I think it's okay to have to pass in initial state as well. If there ends up being a way around that in the future it'll just be a minor version bump, no big deal.

This looks really good to me, we should start moving toward a release. The only thing I want to insist on is that I want a single `effect(() => somePromise)` function rather than `Effects.fromLazyPromise()`. Is it possible to expose the type of the `Effects` class without actually exposing the class itself? To me the existence of the class specifically is an implementation detail.

bdwain:
i think providing an example combineReducers could be useful, but I think most people should write their own because redux-loop can't really optimize it for a general use case. see this old comment I made. https://github.com/redux-loop/redux-loop/issues/51

gyzerok:
Actually I am now thinking that maybe it would make sense to rename `createStore` to something like `createLoopStore` so it would be early on flag for people that they are not *exactly* the same. Any thoughts?

@lukewestby as for `Effects.fromLazyPromise` to `effects` I can easily do it. Before it I would like to show you some things to consider.

 * It feels like it's a bit more clear to see it as class because it do have instance method `map`.
 * As far as I remember previously it was common practise in RxJS, ImmutableJS and maybe some other libraries to have similar API.

Actually these were reasons for me to make it this way. I am not sure if we can expose type without class, but will look into it.

I guess the only thing I need to do in order to make a release is to update examples with something. it should one evening of work.

lukewestby:
@gyzerok +1 for `createLoopStore`, it reminds me of a recommendation that dan abramov made for redux-loop to help people who were having trouble getting the order of the enhancers correct. If it turns out to be really tough to expose the `Effects` type without exposing the class then I'm okay with `Effects.fromLazyPromise`. No point in worrying about names if the work to make them happen introduces unnecessary complexity. This is great! I'm excited to see this moving forward.

lukewestby:
@gyzerok i've had a little change of heart about being able to pass in the arguments to the promise creator so deep-equals testing can be done. I think we should keep it, and if people don't want to use it they don't have to. I've been talking to more people and it sounds like this feature is a major draw for them.

You don't have to implement it before we merge this I'll be happy to spend the time on it since I'm the one creating the idea churn, but I just wanted to let you first instead of committing a bunch of changes later on without letting you know why.

lukewestby:
@gyzerok do you think once this example is fully working this it's okay to merge?

yiransheng:
This looks great! `effect.map` certainly reads a lot more natural compared to previous `Effects.lift`. However to second @lukewestby, exposing `Effects` as class might be a bit two powerful: nothing preventing user doing `effect.toPromise()` inside a reducer. Furthermore, while in typescript land, static types will ensure reducer actually returns `Loop`, when used in javascript, people can simply extend `Effect` class or fake effect object entirely. Previously `getEffect`, `isLoop` etc uses Symbols to detect and extract effect from loop, and the symbol is entirely private to the library, and I am confident a value will never be treated as an effect without calling `Effects.*` apis. Now `getEffect` is simply `ret.effects`, which a rougue reducer could easily fake.

I am not saying class based api is bad, in fact, be able to extend `Effects` could allow a lot of neat tricks - just it does have more potiential for weird or unexpected corner cases.

Another comparision with redux-loop@2.0's, seems with new api, it's very difficult to replicate `Effects.batch`. I do a lot of things like:
```
loop(state,
Effects.batch([
  Effects.promise(fetchData),
  Effects.constant(releaseLoadingLock())
]));
```
Relying on `batch`'s `Promise.all` behavior.


With new api, the following code does something quite different:
```
return {
  state,
  effects : [
    Effect.fromLazyPromise(fetchData),
    Effect.fromLazyPromise(() => Promise.resolve(releaseLoadingLock())
  ]
}
```
To achieve the `batch / Promise.all` behavior, it seems something like this is needed - specifically a new action type on `redux` side to handle an list of actions:
```
return {
  state,
  effects : Effects.fromLazyPromise(() => {
    return Promise.all([fetchData], Promise.resolve(releaseLoadingLock())
      .then([action1, action2] => ({ type: BATCH_ACTION, payload: [action1, action2] });
  })
}
```
The reason is new api diretly passes `dispatch / enhancedDispatch` to effect's promise callback, i.e.:
```
effect.toPromise().then(enhancedDispatch)
```
In redux-loop@2.0, I believe, each effect resolves to an array of actions when converted to promise.

Another appoach with new api:
```
// helper function to ensure a promise returning function gets called
// only once, same thing as lodash.once
function cachePromise(lazyPromiseFn) {
  let cached = null;
  return () => {
    if (!cached) {
      cached = lazyPromiseFn();
    }
    return cached;
  };
}
// inside reducer
const fetchEff = Effects.fromLazyPromise(cachePromise(fetchData));
return {
  state,
  effects: [
    fetchEff,
    fetchEff.map(() => releaseLoadingLock())
  ]
}
```
Caching of promise factory is required to avoid running the same effectual code twice. This solution is better than requiring specific redux batch action type - but it's also non-obvious. I guess a more principled approach would be:

```
class EffectsOnce extends Effects {
  constructor(promiseCreator) {
    this._promiseCreator = once(promiseCreator);
  }
}
```
Which goes back to my previous comments on exposing `Effects` class: greater responsibility but also greater power :)

gyzerok:
@lukewestby yes, I guess we can merge it even earlier. Example is nice for the final release of course, but nobody stops us to make a pre-release, so at least some people will try the thing and give us feedback.

Hello @yiransheng and thanks for your feedback!

> exposing Effects as class might be a bit too powerful: nothing preventing user doing effect.toPromise() inside a reducer. Furthermore, while in typescript land, static types will ensure reducer actually returns Loop, when used in javascript, people can simply extend Effect class or fake effect object entirely. Previously getEffect, isLoop etc uses Symbols to detect and extract effect from loop, and the symbol is entirely private to the library, and I am confident a value will never be treated as an effect without calling Effects.* apis. Now getEffect is simply ret.effects, which a rougue reducer could easily fake.

It's JavaScript! You can see and fake easily everything 😄  With the previous API it was possible as well. After all effect was just an object and symbol was it's property. So you can create this object, read symbol from it and bam you can fake effects. You was able to extend them as well by creating new objects manually with same property names and changing implementation.

Also after all I expect people who come to use this library to have certain level of understanding why they want to use it. And I hope that people won't try to do some hell moves.

> Another comparision with redux-loop@2.0's, seems with new api, it's very difficult to replicate Effects.batch. I do a lot of things like:

> ```
> loop(state,
> Effects.batch([
>   Effects.promise(fetchData),
>  Effects.constant(releaseLoadingLock())
> ]));
> ```
> Relying on batch's Promise.all behavior.

> With new api, the following code does something quite different:

I believe you shouldn't rely on implementation details behaviour at any point. `Promise.all` was probably used to reduce number of rerenders. It was purely optimisation technique. Correct me @lukewestby  if I am wrong. On the other hand it has it's own drowbacks. No you have to wait until the slowest effects get resolved to be able to dispatch all of them.

In your example you should not create two effects immediately, but instead when you recieve actio

```ts
switch (action) {
  case SOMETHING_HAPPEND:
    return {
      state: // something
      effects: [Effect.fromLazyPromise(fetchData)] // this produce message DATA_FETCHED
    }
  case DATA_FETCHED:
    return {
      state: // something
      effects: [Effect.fromLazyPromise(() => Promise.resolve(releaseLoadingLock())]
    }
}
```

yiransheng:
Thanks @gyzerok for the reply. While in general I agree effects should be small and specialized, I do think `Effects.batch` is useful. Even if the library does not support it natively (requiring built-in `Promiss.all` somewhere), I think it should be made eazy to implement on top of `redux-loop`. As I pointed out, this is actually **feasible** with new api. I'd like to add some more thoughts here on the differences between old and new apis, and some quite notes on how to use new api in a backward compatible way.

@lukewestby can probably explain this much better. Currently redux loop `Effects` always resolve to a list of actions. Effects from promise factories have the resolved value automatically wrapped in an array of a single action:

```
Effects.none() -> Promise.resolve([]);
Effects.constant(action) -> Promise.resolve([action]);
Effects.promise(factory) -> factory().then(action => [action]);
Effects.batch(effects) -> Promise.all(effects.map(effectToPromise)).then(flatten)
```

I think this behavior while subtle, outlines a key difference between old and new apis. Previously, a reducer returns a state and a **single** effect, marking `Effects.batch` an essential piece of combining effects. With new api, we are returning a list of effects - and there's no syncrounization on these effects at the library level, any effect's action is dispatched as soon as they are complete. Based on my understanding here's a mapping between old concepts and new ones:

```
Effects.none() -> { state, effects: [] }
Effects.constant(action) ->
  { state, effects: [Effects.fromLazyPromise(() => promise.resolve(action))] } 
Effects.promise(factory) -> { state, effects: [Effects.fromLazyPromise(factory)] }
Effects.batch(effects) -> /* a bit more involved */
```

With this in mind, a compatibility layer can be established:

```
// redux-loop-compat.js

export function loop(state, effects) {
  if (!Array.isArray(effects)) {
    effects = [effects];
  }
  return { state, effects };
}

// all apis return an single effect or an array of effects
const EffectsApi = {
  none : () => [],
  constant : (action) => Effects.fromLazyPromise(() => Promise.resolve(action)),
  promise : (factory, ...args) => Effects.fromLazyPromise(() => factory(...args)),
  lift : (effect, factory) => effect.map(factory),
  batch: (effects) => {
    const factoryForAll = once(() => {
      const promises = effects.map(e => e.toPromise());
      return Promise.all(promises);
    });
    const batchEff = Effects.fromLazyPromise(factoryForAll);
    // returns an array of effects
    return effects.map((_, i) => batchEff.map(actions => actions[i]));
  }
};
export { EffectsApi as Effects };
```

Then in an existing code base, just needs to change import statements, deprecate all `Effects.call` usages.

```
import {loop, Effects} from "redux-loop-compat";

// old reducer code remains the same
// new reducers can adopt new api
```

lukewestby:
@gyzerok i have some time today so i'm gonna merge this and keep working on things

gyzerok:
@lukewestby ok, thanks! Sorry for a bit of an off time, have some release to be done soon :(

PR #111, opened 2017-03-07
Fixes liftState and compose example.

artemruts:
Hey,

I think when using `compose` from `redux` and `liftState`, we should make sure `reducer` is the rightmost function to maintain the `(state, action)` signature.

https://github.com/reactjs/redux/blob/master/src/compose.js#L1:L10

Thanks!


bdwain:
Hi. Sorry for the lack of response. 

I have some significant changes coming up so I don't think this change will apply anymore. But thank you for contributing.

PR #114, opened 2017-03-17
add transform runtime

thewillhuang:
I've been wanting to convert my company's codebase to stop using polyfill, saving some bandwidth, however for IE11, I run into error regarding Symbol being undefined from redux-loop.

This PR makes redux loop use transform which should be best practice for a library.

see https://babeljs.io/docs/plugins/transform-runtime/

> Another purpose of this transformer is to create a sandboxed environment for your code. If you use babel-polyfill and the built-ins it provides such as Promise, Set and Map, those will pollute the global scope. While this might be ok for an app or a command line tool, **it becomes a problem if your code is a library which you intend to publish for others to use or if you can’t exactly control the environment in which your code will run.**

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,6 +3,7 @@
   "plugins": [
     "transform-object-rest-spread",
     "transform-es3-member-expression-literals",
-    "transform-es3-property-literals"
+    "transform-es3-property-literals",
+    "transform-runtime"
   ]
 }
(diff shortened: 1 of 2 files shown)


lukewestby:
We switched the build entirely to the typescript compiler so this won't be needed anymore. Thanks for putting in the effort though, sorry we couldn't accept it!

thewillhuang:
awesome, thank you!

PR #132, opened 2017-07-15
add backwards compatible effects until next major version

bdwain:
these will make it easier for people to transition to the new Cmd style api slowly rather than all at once. We should keep them around until version 4.0.

I didn't think it was worth adding the other deprecated effects because they were all just renamed and so it was just find/replace to update them to the new api.

### DIFF
--- a/modules/cmd.js
+++ b/modules/cmd.js
@@ -88,7 +88,9 @@ export const cmdToPromise = (cmd, dispatch, getState) => {
     case cmdTypes.MAP:
       const possiblePromise = cmdToPromise(cmd.nestedCmd, dispatch, getState)
       if (!possiblePromise) return null
-      return possiblePromise.then((actions) => actions.map(cmd.tagger))
+      return possiblePromise.then((actions) => 
+        actions.map(action => cmd.tagger(...cmd.args, action))
+      );
 
     /*case cmdTypes.CALLBACK:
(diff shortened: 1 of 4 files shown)


ashishsc:
If we can't backport all of the effects, won't the client code crash anyway if those functions are called? My point is, is there value in only offering just a few of the lib functions, or would that be more confusing? I could go either way, but the the code lgtm if we want to merge it.

bdwain:
my thought was that since promise and call can't be trivially updated to Cmd.run, it would take more time and the backwards compatibility would be more useful. But everything else is just find and replace, and there's not much good reason to delay fixing it. I can add the rest though if you think it could be confusing. I was torn on it anyway.

modules/cmd.js
```diff
@@ -88,7 +88,9 @@ export const cmdToPromise = (cmd, dispatch, getState) => {
     case cmdTypes.MAP:
       const possiblePromise = cmdToPromise(cmd.nestedCmd, dispatch, getState)
       if (!possiblePromise) return null
-      return possiblePromise.then((actions) => actions.map(cmd.tagger))
+      return possiblePromise.then((actions) =>
```
bdwain: noticed that the implementation of Cmd.map didn't actually take the args

ashishsc:
Don't forget to squash

PR #135, opened 2017-07-30
Fix overzealous error message

IamfromSpace:
This fixes an issue where loop considered errors after dispatch to be errors in the effect, which provides a misleading error.

As an example, a state update may cause an error in the view.  If this update came as the result of an effect, that error would be caught and report a loop error, because the catch comes after the dispatch.  However, the effect worked as intended, and the error was down stream.

By moving the catch to immediately after the effect promise, we only catch errors that are occur in the effect.

This patch back ports to all current minor releases, and the most recent previous major/non-dev releases (as long as it is not squashed on merge).

87d9e24 can be immediately tagged and released as v1.1.1
ddf36c7 can be immediately tagged and released as v2.2.3
9449abb can be immediately tagged and released as v3.0.1
bf6f15b (HEAD) can be immediately tagged and released as 3.1.1

All tests pass for all four versions.


modules/install.js
```diff
@@ -38,14 +38,14 @@ export function install() {
       if (!cmdPromise) return null
 
       return cmdPromise
-        .then((actions) => {
```
bdwain: one issue with this is that errors in the action creators (not dispatching them, just creating them) would not generate a loop error, and I'd argue they should.

Maybe the solution is to generate the actions first, then catch, then dispatch after?

bdwain:
Thanks! I like the idea of not making errors due to dispatching actions look like loop errors. I want to make sure those errors don't get swallowed though. They will show up in the console still, right?

One other thing is that while I appreciate you backporting the change, the only version we need to support is the latest one. It is backwards compatible with v2, and v1 is very old and no longer supported. So can you only commit the one commit?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "redux-loop",
-  "version": "3.1.0",
+  "version": "3.1.1",
```
bdwain: can you remove the version bump please? i can bump and tag master after this merged.

IamfromSpace: Yep, can do.  For the back-ported versions, is it okay to leave them?  The backporting strategy I've used automatically incorporates each new version into history, which would be lost if they were bumped after merge.  If you want, I can separate out the bumps into separate commits.

modules/install.js
IamfromSpace: This does actually catch errors within the action creator, because the cmdPromise is the promise for the actions.  If an error occurs while the cmdPromise is generating the actions, the new position of catch will still log it--which is exactly what we're interested in catching/logging.  Once we have the actions and are ready to dispatch, we want to be hands off and let errors go uncaught.

IamfromSpace:
No risk of swallowing the errors, the only behavior that will change is what gets logged.

**If the effect throws:**
_Master_: it will skip the then/dispatch, be caught, logged, and re-thrown, and be uncaught.
_Patch_: it will be caught, logged, re-thrown, skip the then/dispatch, and be uncaught.

**If post-dispatch throws:**
_Master_: it will dispatch, be caught, logged, re-thrown, and then uncaught.
_Patch_: it will dispatch, and then be uncaught.

As for the back-porting, I understand that v1 is old, and I'm happy to squash to remove v1.  However, I'm still using v2, and am very interested in a patch for that version.

package.json
bdwain: does your app break if you upgrade to the latest v3 (3.1)? I think it should be backwards compatibile. 3.0 was breaking but 3.1 brought back compatiblity with v2 (will be removed in v4).  If it's not actually backwards compatible  then we can do v2 also. if it is though,, then we only need 3.1

modules/install.js
bdwain: ah good point i missed that.

package.json
bdwain: see the other comment. but since 3.1 is supposed to be backwards compatible with v2, i think we only need the latest. if you have trouble upgrading to v2 though, then we can do it. it's supposed to work though.

IamfromSpace:
Got some updated commits that try to address as much as possible:

I tried to migrate to 3.1.0, but I had some trouble because `getEffect` is no longer available, and I've got some loop dependent util libraries that are spread out.

However, I removed the 3.1.1 bump, and removed all back ports except for 2.2.2, and made the version bump a separate commit to match the style of the repo.  If you're willing to publish that commit I'd be very appreciative!

I also added some tests to validate that the error catching does act as intended (partly to double check that I wasn't crazy!) and I did come across an issue where errors could be swallowed, so I changed the default errorActionCreator to be undefined.

The reason here is that while it make sense that we might want to emit a side effect that we don't want to know the result of (I've come across this a number of times), we _probably_ want to know about any errors that occur.

If the user doesn't define an errorActionCreator, or simply misspells it, the behavior on master will catch, return undefined, and then filter out the result.  With JS it's pretty easy to accidentally end up with an expected undefined, and then think that errors aren't occurring when they are.

If you can pull down the branch you can validate the error get's consumed by adding in the original default function and then running the tests.

Let me know if you have any more comments or concerns!

bdwain:
Sorry i didn't realize i forgot to add getEffect back. I released 3.1.1 with getEffect in it. I also fixed an issue that made it hard to compare the effects from your reducers with effects in your tests.

I believe it should be fully backwards compatible now. Would you mind trying out the new version? I'd rather focus on making v2 unnecessary instead of improving it unless there are big changes needed to make v3 work with old code.

bdwain:
hi @IamfromSpace. did the latest v3 work for you?

bdwain:
I'm going to close this for now. If you want to revisit it we can reopen.

PR #139, opened 2017-08-08
Import isCmd into Effects module to solve an issue where it was undefined.

IamfromSpace:


bdwain:
thanks for catching this!

Issue #130, opened 2017-07-12
Upcoming Changes


bdwain:
Sorry for the long delay between updates everyone. 

I released a version 3.0 that should solve some of the problems with version 2. You can see the details [here](https://github.com/redux-loop/redux-loop/releases/tag/v3.0.0). It's not a huge departure from version 2, and I think it should set us up to make more improvements to the project.

There's still a few things I would like to address quickly before working on new features.
- [x] v2 backwards compatibility
- [x] Typescript Typings (if anyone has any experience with this and is interested in helping, that would be appreciated)
- [x] Rewrite tests in jasmine or mocha
- [x] Update build tasks

After those are done, I think we'll be in a good position to start improving the API more based on feedback from people who are using it.

ashishsc:
I think the question about TS support is whether we should just define typings or rewrite the library in typescript itself. One is larger endeavor than the other obviously.

bdwain:
I think for now it'd be good to just do typings since that seems to be in high demand and I'd rather not keep people waiting anymore. 

I think it's worth discussing if we want to rewrite everything in typescript. I'm not sure I see a big benefit from it. But I don't have much experience with it so I'd be interested in hearing other peoples' opinions.

ashishsc:
I agree that we should just do the typings in the short term. The benefit of a typescript rewrite are the same as any project: it eliminates having to

- maintain a seperate typings file (I would hope)
- write classes of unit tests checking for the correct type since the compiler will catch these

And in addition, it will make it easier for us to
- refactor the codebase in the future
- have new developers come on (reading the type signatures is helpful, same as jsdoc)

bdwain:
that makes sense. well i guess if we are going to rewrite the tests and build process anyway, may as well do full typescript conversion before we do those.

ashishsc:
For reference:
https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html

I'm going to read this tomorrow

gyzerok:
There was already new API fully rewritten in TypeScript. I suggest to use it, because it simplifies library quite a lot and got a lot of thought in it. It was planned as 4.0.

MazeChaZer:
@gyzerok I am currently using that API in my TypeScript project, works great :+1:

bdwain:
[...]

PR #141, opened 2017-08-30
Add Type Declarations For Redux-Loop V3

willisplummer:
This is a first pass at type declarations for `redux-loop`. So far, I've added all of the documented types of `Cmd`, `install` and `combineReducers` but not any of the deprecated old API.

If it seems necessary to add that stuff, it's definitely possible, but it'll complicate some of the types.

I've also fixed a few typos in the README.

A compiling example of a typescript project using these definitions [can be found here](https://github.com/willisplummer/redux-loop-ts)


willisplummer:
👋 @bdwain re #130

bdwain:
thanks for posting this! It'll take me a few days to look because I don't have much experience with typescript but I'll respond soon.

bdwain:
can you add a field to package.json `"types": "./lib/main.d.ts"`

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+
+interface EnhancedReducer<S, T> {
+  <A extends Action>(state: S, action: A): T;
+}
+
+interface NoneCmd {
+  type: 'NONE';
+}
+
+interface ActionCmd<A> {
```
bdwain: should A extend Action?

index.d.ts
```diff
@@ -0,0 +1,78 @@
+import { Action, StoreEnhancer, Store } from 'redux';
+
+interface StoreCreator {
```
bdwain: should all of these interfaces and types be exported? they are in redux's typings file. https://github.com/reactjs/redux/blob/master/index.d.ts

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+interface ActionCmd<A> {
+  type: 'ACTION';
+  actionToDispatch: A;
+}
+
+interface BatchCmd<AppAction extends Action, F extends Function> {
+  type: 'BATCH';
+  cmds: CmdType<AppAction, F>;
```
bdwain: should this be `CmdType<AppAction, F>[]` since it's an array of them?

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+  options: {
+    args: any[];
+    failActionCreator: Function;
+    successActionCreator: Function;
+  };
+}
+
+interface SequenceCmd<AppAction extends Action, F extends Function> {
+  type: 'Sequence';
+  cmds: CmdType<AppAction, F>;
```
bdwain: same here with the array

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+
+interface StoreCreator {
+  <S, T>(
+    reducer: EnhancedReducer<S, T>,
+    preloadedState: S,
+    enhancer: StoreEnhancer<S>
+  ): Store<S>;
+}
+
+interface EnhancedReducer<S, T> {
```
bdwain: the reducer type in redux is just a function type.

`export type Reducer<S> = (state: S, action: AnyAction) => S;`

can this not be one also?

also, I think we know that T is either an S or a loop(S, CmdType) right? Can we restrict it to that?

bdwain: ah i see you can user interfaces the same way as the type itself. but the last question still applies

index.d.ts
willisplummer: 👍

index.d.ts
willisplummer: 👍  not sure how these slipped by 😓

index.d.ts
willisplummer: 👍

index.d.ts
willisplummer: 👌  yeah seems good to just emulate what they're doing in the redux declaration

index.d.ts
willisplummer: One instance where `EnhancedReducer` gets a different return type via the generic is in `combineReducers` where we know that the response is always going to be a `loop(S, CmdType)` and never just State:

```
declare function combineReducers<S, T>(
  reducers: StrictReducerMapObject<T>
): EnhancedReducer<S, [S, CmdType<Action, Function>]>;
```

Not sure if that's worth adding an extra generic for though? Maybe there's a more concise way of expressing that in this case we know which member of the union type will be returned.

index.d.ts
```diff
@@ -0,0 +1,82 @@
...
+    preloadedState: S,
+    enhancer: StoreEnhancer<S>
+  ): Store<S>;
+}
+
+export interface EnhancedReducer<S> {
+  <A extends Action>(state: S, action: A): S | [S, CmdType<A, Function>];
+}
+
+declare interface LiftedEnhancedReducer<S> {
```
willisplummer: @bdwain okay - I got rid of the second generic argument to EnhancedReducer and then created LiftedEnhancedReducer to reflect cases like combineReducers where the return type will never just be `S` -- might be a more concise way of expressing this, but for a first pass this should work

index.d.ts
```diff
@@ -0,0 +1,82 @@
...
+export interface StoreCreator {
+  <S, T>(
+    reducer: EnhancedReducer<S>,
+    preloadedState: S,
+    enhancer: StoreEnhancer<S>
+  ): Store<S>;
+}
+
+export interface EnhancedReducer<S> {
+  <A extends Action>(state: S, action: A): S | [S, CmdType<A, Function>];
```
bdwain: maybe we can extract loop into its own type. People will need that

index.d.ts
bdwain: good idea!

index.d.ts
```diff
@@ -0,0 +1,82 @@
...
+
+declare interface LiftedEnhancedReducer<S> {
+  <A extends Action>(state: S, action: A): [S, CmdType<A, Function>];
+}
+
+export interface NoneCmd {
+  type: "NONE";
+}
+
+export interface ActionCmd<A extends Action> {
```
bdwain: what's the reason for making all of the Cmds generics? I get the reducer because we want to make sure the function always returns the right type. But won't you just say something like

```js
let cmd: ActionCmd = Cmd.action(someAction)
return loop(newState, cmd);
```

in the reducer?

index.d.ts
```diff
@@ -0,0 +1,82 @@
+import { Action, StoreEnhancer, Store } from "redux";
+
+export interface StoreCreator {
+  <S, T>(
+    reducer: EnhancedReducer<S>,
```
bdwain: what are your thoughts on renaming EnhancedReducer to LoopReducer? I feel like enhanced is more generic than it needs to be.

index.d.ts
```diff
@@ -0,0 +1,82 @@
...
+  | BatchCmd<AppAction, F>
+  | NoneCmd
+  | RunCmd<F>
+  | SequenceCmd<AppAction, F>;
+
+declare function install<AppState>(): StoreEnhancer<AppState>;
+
+declare function loop<State, AppAction extends Action, F extends Function>(
+  state: State,
+  effect: CmdType<AppAction, F>
```
bdwain: this should be named cmd

bdwain:
@willisplummer there's a few things from the API that are not included that I think would be good to have.

- isLoop
- getModel
- getCmd
- liftState
- Cmd.map
- Cmd.dispatch
- Cmd.getState

None of those are deprecated so I'd like to support them too. Let me know if any of those cause issues.

index.d.ts
willisplummer: 👍 sgtm

index.d.ts
willisplummer: not sure i'm understanding but the loop function is declared on line 57 

https://github.com/willisplummer/redux-loop/blob/96eb5a6ecc7921f9c1b9f2f1d16a68f0dc77152e/index.d.ts#L57

index.d.ts
willisplummer: 👍

index.d.ts
bdwain: im talking about the type of the return value of loop(), which is `[S, CmdType]`. That's returned by loop(), and also by loopreducers

willisplummer: got it - added that

index.d.ts
willisplummer: Good question. I'm starting to think that the `F` generic for the `CmdType` isn't necessary. 

The thinking behind making Action a generic is that in the reducer file you can define a union type of Actions that your reducer expects to handle and then you get an additional layer of typesafety ensuring that your Cmd.action() will be receiving actions that the reducer knows how to handle.

i'd recommend taking a look at the [sample project here](https://github.com/willisplummer/redux-loop-ts/blob/master/src/index.tsx#L8-L19).

for example:

```
type Action = { type: 'INCREMENT', amount: number } | { type: 'INCREMENT_AGAIN' }

const reducer: EnhancedReducer<
  number,
  number | [number, CmdType<Action>]
> = (state = 0, action: Action) => {
  switch (action.type) {
    case 'INCREMENT':
      return loop(state + amount, Cmd.action({type: 'INCREMENT_AGAIN'})
    case 'INCREMENT_AGAIN':
      return state + 1;
    default:
      return state;
  }
}
```

Without the generic if you accidentally wrote a typo like `return loop(state + amount, Cmd.action({ type: 'INCREMENT_AGAN'})` that would be a valid `Cmd.action` because it is a redux `Action` (an object with a `type` property) but your reducer would just fall through to the default case and nothing would happen.  Similarly, if you tried to call `Cmd.action({ type: 'INCREMENT', amount: 'not a number' })` typescript wouldn't be able to catch that for you.

willisplummer:
👍 i'll take a look at adding the rest of these over the weekend

index.d.ts
```diff
@@ -0,0 +1,84 @@
...
+  type: 'ACTION';
+  actionToDispatch: A;
+}
+
+export interface BatchCmd<A extends Action> {
+  type: 'BATCH';
+  cmds: CmdType<A>[];
+}
+
+type ActionCreator<A extends Action> = (...args: any[]) => A;
```
bdwain: redux has an ActionCreator type that we should be able to reuse

index.d.ts
bdwain: but all reducers get called with all actions anyway. So the reducer can't limit what actions it receives. it just needs to know to use the default handler for anything it doesn't recognize.

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+  <S, T>(
+    reducer: LoopReducer<S>,
+    preloadedState: S,
+    enhancer: StoreEnhancer<S>
+  ): Store<S>;
+}
+
+export type Loop<S, A extends Action> = [S, CmdType<A>];
+
+export interface LoopReducer<S> {
```
bdwain: is there a difference between this current syntax and putting both generic params on the interface?

```js
export interface LoopReducer<S, A extends Action> {
  (state: S, action: A): S | Loop<S, A>;
}
```

bdwain:
sweet! this is shaping up nicely. Thanks for helping out on this!

test/typescript.test.js
```diff
@@ -0,0 +1,10 @@
+import test from "tape";
```
willisplummer: @bdwain here i followed the pattern used by redux for testing their type declarations. seemed safe to introduce the same dependencies as them + beneficial to have some tests for the purposes of refining the type system in less of a vacuum

index.d.ts
```diff
@@ -0,0 +1,103 @@
...
+  reducers: StrictReducerMapObject<T, A>
+): LiftedLoopReducer<S, A>;
+
+declare function liftState<S, A extends Action>(
+  state: S | Loop<S, A>
+): Loop<S, A>;
+
+declare function isLoop(test: any): boolean;
+
+declare function getModel(a: any): any;
```
willisplummer: from my understanding of `getModel`, it's not really possible to get any type safety around it

index.d.ts
willisplummer: yeah - It's that the types add a guard against what actions the commands can dispatch

willisplummer:
👋 @bdwain just wanted to follow up to see if you've had a chance to take a look at the latest change set? i think we're in pretty good shape and there's definitely room to refine and tighten the type system once it's in place and people are using it in context(?)

bdwain:
@willisplummer sorry not yet i was out of town this weekend. I'll check it out tonight though.

index.d.ts
```diff
@@ -0,0 +1,110 @@
...
+    reducer: LoopReducer<S, A>,
+    preloadedState: S,
+    enhancer: StoreEnhancer<S>
+  ): Store<S>;
+}
+
+export type Loop<S, A extends Action> = [S, CmdType<A>];
+
+export interface LoopReducer<S, A extends Action> {
+  (state: S | undefined, action: A): S | Loop<S, A>;
```
bdwain: @willisplummer where i'm still confused about the generics on the Action types is here. A is used not just in the return type, but in the parameter for the Action. But all reducers need to be able to take any action type, not just a subset. Should this line be `(state: S | undefined, action: AnyAction): S | Loop<S, A>;`

index.d.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+declare interface LiftedLoopReducer<S, A extends Action> {
```
bdwain: why is this declared but LoopReducer is exported?

willisplummer: ah yeah this should probably be exported too

index.d.ts
bdwain: should it not be the same thing as loop() ?

test/typescript.test.js
bdwain: sweet looks awesome!

index.d.ts
```diff
@@ -0,0 +1,110 @@
...
+declare function loop<S, A extends Action>(
+  state: S,
+  loop: CmdType<A>
+): Loop<S, A>;
+
+declare class Cmd {
+  static readonly none: NoneCmd;
+  static action: <A extends Action>(action: A) => ActionCmd<A>;
+  static batch: <A extends Action>(cmds: CmdType<A>[]) => BatchCmd<A>;
+  static readonly dispatch: (action: Action) => void;
```
bdwain: just for readability can you move dispatch and getState up above or below all of the Cmd.foo functions that generate commands.

index.d.ts
```diff
@@ -0,0 +1,110 @@
...
+declare function install<S>(): StoreEnhancer<S>;
+
+declare function loop<S, A extends Action>(
+  state: S,
+  loop: CmdType<A>
+): Loop<S, A>;
+
+declare class Cmd {
+  static readonly none: NoneCmd;
+  static action: <A extends Action>(action: A) => ActionCmd<A>;
```
bdwain: why not make these readonly like the other ones?

tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+    "module": "commonjs",
+    "paths": {
+      "*": ["*"]
+    },
+    "sourceMap": false,
+    "strict": true,
+    "target": "es6",
+    "typeRoots": ["./node_modules/@types", "./index.d.ts"]
+  },
+  "include": ["./lib/**/*", "./test/**/*.ts"]
```
bdwain: should this be the module folder instead of the lib folder? we shouldn't need to build before running tests right?

index.d.ts
willisplummer: @bdwain that makes sense to only include the generic in the return type I think.

I think this stackexchange question was the origin of this typing and it seems like we still get the safety I was after with `action: AnyAction`

https://stackoverflow.com/questions/45899039/how-do-i-write-a-function-type-that-strictly-requires-its-parameters-in-typescri/45900816#45900816

Nice suggestion 👍

index.d.ts
```diff
@@ -0,0 +1,110 @@
...
+    options?: {
+      args?: any[];
+      failureActionCreator?: ActionCreator<A>;
+      successActionCreator?: ActionCreator<A>;
+    }
+  ) => RunCmd<A>;
+  static sequence: <A extends Action>(cmds: CmdType<A>[]) => SequenceCmd<A>;
+}
+
+export interface StrictReducerMapObject<T, A extends Action> {
```
bdwain: could this just be ReducerMapObject? Strict makes it sound like there's a non-strict version

index.d.ts
willisplummer: 👍

index.d.ts
willisplummer: 👍

index.d.ts
willisplummer: hmm i was using `readonly` to indicate that these were only getter attributes. i guess it's true that it could also be applied to the class methods and doesn't appear to break tests so 👍

tsconfig.json
willisplummer: that makes sense!

willisplummer:
@bdwain made the requested changes if everything seems okay to you?

index.d.ts
willisplummer: @bdwain maybe... I think I may have misunderstood the docs at first. does this look right to you?

```
declare function getModel<S>(a: Loop<S, AnyAction> | S): S;
```

bdwain:
@willisplummer Everything is looking great. A few final things and I think this will be good to merge.

Can you add index.d.ts to the files array in package.json? Without that, I don't think the typings file will be in the published package.

Then, can you rebase your branch off of the latest master and squash all of the commits down to a single commit? No need to retain all of the individual commit messages. The description can just be something like "add typescript typings".

Thanks for all of your hard work on this!

willisplummer:
@bdwain should be good to go - can you just take a look at the package.json files array and make sure i did that right.. wasn't sure if i needed to specify a relative path

bdwain:
looks good! that's how redux does it so i assume it's fine

bdwain:
thanks again @willisplummer ! This is awesome!

bdwain:
this was published in 3.2.0

PR #142, opened 2017-09-02
Spread old state when creating new state

tschaub:
This updates the API docs so that examples spread the old state (`...state`) when returning new state from a reducer.


bdwain:
Thanks for fixing that!

PR #148, opened 2017-09-21
Rewrite tests with jest

bdwain:
this is the last thing I'd like to do before releasing 4.0 and then we can start working on new features.

@ashishsc @willisplummer if either of you guys could give this a look It'd be good to get a second set of eyes on it. 

Thanks!


src/cmd.js
```diff
@@ -15,7 +14,7 @@ const cmdTypes = {
 }
 
 export const isCmd = (object) => {
-  return object ? object[isCmdSymbol] : false
+  return object ? !!object[isCmdSymbol] : false
```
willisplummer: curious about this change?

src/loop.js
```diff
@@ -1,12 +1,10 @@
 import { throwInvariant } from './utils';
 import Cmd, { isCmd } from './cmd';
 
-
 export const isLoop = (array) => {
-  return Array.isArray(array) && array.length === 2 && isCmd(array[1])
+  return Array.isArray(array) && array.length === 2 && isCmd(array[1]) && !isCmd(array[0])
```
willisplummer: 👍

test/combineReducers.spec.js
```diff
@@ -0,0 +1,78 @@
...
+import { Map } from 'immutable';
+
+const reducers = {
+  counter: (state = 0, action = {}) => state + 1,
+  doubler: (state, action = {}) => state ? state + state : 1,
+  fibonacci: (state = 1, action = {}) => action.previous ? action.previous + state : state
+};
+
+describe('combineReducers', () => {
+  it('works with one argument and returns correctly working reducer', () => {
```
willisplummer: so one thing that i've found in working with jest is that making it statements apply to as few `expect`s as possible makes debugging test failures a lot more pleasant - otherwise it can get hairy pretty quickly to figure out which of the assertions is failing. just something to keep in mind as we write tests or attempt to debug test failures

test/utils.spec.js
```diff
@@ -0,0 +1,38 @@
...
+      expect(flatten(arr)).toEqual([1, 2, 3, 4, 5, 6]);
+    });
+  });
+
+  describe('throwInvariant', () => {
+    it('throws an error with the supplied message if the condition is false', () => {
+      expect(() => throwInvariant(false, 'sad panda')).toThrow(new Error('sad panda'));
+    });
+
+    it('does not throw an error if the condition is false', () => {
```
willisplummer: if the condition is true

test/utils.spec.js
```diff
@@ -0,0 +1,38 @@
+import {flatten, throwInvariant, isPromiseLike} from '../src/utils';
```
willisplummer: do we want to add tests for `mapValues` and `promisify`?

test/index.spec.js
```diff
@@ -0,0 +1,117 @@
+import { install, loop, Cmd, combineReducers } from '../src';
+import { createStore } from 'redux';
+
+describe('redux loop store enhancer', () => {
+  it('a looped action gets dispatched after the action that initiated it is reduced', done => {
```
willisplummer: why are we doing these tests with a combined reducer? it seems like unnecessary additional complexity

src/cmd.js
bdwain: when an object was not a cmd i just wanted it to return false instead of undefined.

test/combineReducers.spec.js
bdwain: yea i agree. the combineReducers file and the index file though i just took the existing tests and converted them to jest. maybe at a later point we can break up the tests more.

test/utils.spec.js
bdwain: oops

test/utils.spec.js
bdwain: those have been removed in this branch (really in this 4.0 branch, which this branched off of)

test/index.spec.js
bdwain: see above. just didn't want to rework all of these tests from scratch for now. that'll be another project. just got them running on jest for now.

willisplummer: 👍

ashishsc:
sorry I was recruiting in Michigan!

PR #150, opened 2017-09-22
fix error with Cmd.run options typing

gniquil:


bdwain:
thanks for fixing that!

Issue #133, opened 2017-07-17
5.0 Ideas


bdwain:
The Cmd API released in version 3 was the first attempt at simplifying the API and making effects more reusable. But there are still more improvements that can be made. Let's use this thread to post feedback on the current version and ways to improve it. 

- New Cmd types like batch and sequence, except they dispatch resulting actions as they become ready instead of all at once at the end.
- Break out integrations with popular libraries into separate projects (such as immutablejs-optimized combineReducers)
- Expose a method to help nest reducers (like combineReducers but where the common parent has functionality of its own).

icopp:
I'd personally propose replacing the `Cmd.x` functions with individual exports with names like `queuePromise` and `queueDispatch`. Similarly, `Cmd.getState` and `Cmd.dispatch` would probably be better as something like `getStatePlaceholder` and `dispatchPlaceholder`.

There's also the issue that `loop` as a name for the function to return is kind of weird, since in the common parlance it's not actually looping anything; that's what got me thinking of names using terminology like `queue` in the first place.

ashishsc:
I guess at some point we should decide if we want to try and mimic elm's function names or not. Task and Effect? `thunk` ? `schedule` ?

icopp:
"Thunk" seems like a bad idea, given that in this space it tends to mean `redux-thunk` pretty specifically.

bdwain:
i'm not married to any particular name, but i'd want to not change them without evidence that Cmd was confusing people or that a new name would be much more clear

farism:
I like having things be the same as elm. I kind of think of redux-loop as an elm polyfill for redux and should strive to match the implementation as closely as possible. With additional features as necessary, of course.

That said, configurable batching behavior would probably be at the top of the list in terms of features I would like to see. A parallel batch would be nice, which @bdwain already mentioned in the first comment.

bdwain:
Version 4.0 was just released. It was a small update to remove deprecated functionality and clean up the build and tests to make future changes easier. 

Now that that's done, I'd love to start tackling some of these ideas and coming up with more. 

The first one I'd like to do is configurable batching behavior. My thought is that we have batch and sequence now, but it'd probably be simplest for there to be one Cmd type for an array of Cmds, with 2 options available. Option 1 is a boolean to choose batch vs sequence (parallel vs series) behavior. Option 2 is to determine whether all actions dispatch at once at the end (the current behavior) or each action gets dispatched as it becomes ready.

[...]

PR #151, opened 2017-09-23
Configurable batching

bdwain:
This adds support for the configurable batching behaivor in #133. It deprecates Cmd.batch and Cmd.sequence in favor of Cmd.list. List takes an array of cmds and an options object. It still needs updates (tests, docs, typescript typings), but I think the code should be good to go and ready to be looked at.

There are 2 options for list: batch and sequence (yes the old cmd names. I think they make the most sense). The sequence option decides between parallel vs series (like Cmd.batch vs Cmd.sequence), and is off by default. The batch option decides whether to dispatch the actions as they are ready or all at once at the end (the current behavior for both). It will be disabled (no batching) by default, which means the default Cmd.list has slightly different behavior than batch/sequence.

```js
Cmd.batch(arr) === Cmd.list(arr, {batch: true})
Cmd.sequence(arr) === Cmd.list(arr, {sequence: true, batch: true})
```

Other names that were considered were for
- Cmd.list: Cmd.group, Cmd.array
- batch: group/groupCmds
- sequence: series

I'm going to add README changes and typings changes and tests still, just wanted to get this initial implementation out there.

This change can be released as part of 4.x and then in 5.0 we can do the breaking portions of the change, which include removing Cmd.batch and Cmd.sequence and changing the default behavior of combineReducers to not batch child cmds.


src/cmd.js
```diff
@@ -49,41 +50,65 @@ function handleRunCmd(cmd, dispatch, getState){
...
+  return Promise.all(promises).then(flatten);
+}
+
+function handleSequenceList({cmds, batch}, dispatch, getState){
+  const firstCmd = cmds.length ? cmds[0] : null;
+  if(!firstCmd){
+    return null;
+  }
+
+  let result = new Promise(resolve => {
```
farism: Maybe use `const result` here?

src/cmd.js
```diff
@@ -49,41 +50,65 @@ function handleRunCmd(cmd, dispatch, getState){
...
+    let firstPromise = executeCmd(firstCmd, dispatch, getState);
+    firstPromise = firstPromise || Promise.resolve([]);
+    firstPromise.then(result => {
+      if(!batch){
+        result.map(dispatch);
```
farism: Maybe use `result.forEach` here?

farism:
The implementation looks good.

Left a few very minor comments.

README.md
```diff
@@ -249,7 +251,7 @@ are using using something other than a plain object as your state, you should
 consider writing your own implementation of combineReducers, using our 
 version as a reference. It's impossible to provide a generic combineReducers
 that will be optimal for all state shapes. As an example, updating an immutable.js
-state would be much faster if you made use of the withMutations method to batch updates. 
+state would be much faster if you made use of the withMutations method to  updates.
```
farism: 'to run updates'?

bdwain: oops good catch

src/cmd.js
```diff
@@ -49,41 +50,67 @@ function handleRunCmd(cmd, dispatch, getState){
...
+    return possiblePromise.then(result => {
+      return Promise.all(result.map(dispatch));
+    });
+  }).filter(x => x);
+
+  if (promises.length === 0){
+    return null;
+  }
+
+  return Promise.all(promises).then(flatten).then(actions => {
```
bdwain: i made an implementation change here. I realized that the promise returned by dispatch should probably still wait for everything to be finished before resolving, even if it doesn't wait to dispatch any actions. Now it resolves at the end with an empty array of actions to dispatch

src/combineReducers.js
```diff
@@ -41,6 +41,6 @@ export const combineReducers = (
 
   return loop(
     hasChanged ? model : state,
-    Cmd.batch(cmds)
+    Cmd.list(cmds, {batch: true}) //todo: remove batch in 5.0
```
bdwain: don't want to change the behavior until a breaking change to avoid catching people off guard.

index.d.ts
```diff
@@ -50,19 +54,20 @@ export interface RunCmd<A extends Action> {
   };
 }
 
-export interface SequenceCmd<A extends Action> {
-  type: 'Sequence';
-  cmds: CmdType<A>[];
-}
+//deprecated types
```
bdwain: @willisplummer do these type changes make sense? Since Cmd.batch and Cmd.sequence will now return instances of Cmd.list, it seemed like we could just make those types aliases of List.

bdwain:
@farism i made some implementation changes since you looked. mostly focused around keeping the behavior that the dispatch method returns a promise that waits for all resulting actions to complete. Now that still works because executePromise resolves with an empty array, but not until the nested actions are all dispatched and processed.

jarekb84:
This PR seems like an attempt to simplify the API by combining two similar commands under one, and using options to disable current behavior. I like the concept of using an options parameter instead of a long series of individual params, some of which maybe optional. But when I see lots of options, some of which are not needed in all cases, that comes off as a code smell indicating a method is doing too many things and needs to be broken up into multiple specific sub methods. We're not at the point where you have tons of options, but they're being used here for divergent needs.

I like the existing names for these actions, they are explicit and fairly intuitive in their purpose. I'm also a fan of a stable api, where core methods do not get renamed or repurposed often or at all. 

My suggestion would be to keep Cmd.Batch and Cmd.Sequence as is, and add an options param to toggle the existing batching of when actions are dispatched. This would be a breaking change if the default is batch=false which would give you the performance benefit of early dispatches.

If you still want to rename Cmd.Batch to something else, what about Cmd.Parallel or Cmd.Independant.

bdwain:
What does everyone else think about @jarekb84 's comment? @ashishsc @farism @willisplummer ?I'll try to summarize the reasons I merged batch and sequence.

- Whether a list of cmds is run in series or in parallel, the concept of running a bunch of Cmds is basically the same. They may completely different implementations under the hood, but from the user's perspective they are very similar. I worry about adding too many methods to the top level API because I think it could confuse newcomers more about which they should use. Run, map, action and none are all very clear and distinct about how they are different, and list (or whatever we call it) would fit in with that group well. Adding 2 very similar methods to the top level API though would make it more confusing IMO.
- In most scenarios, I see parallel unbatched groups (aka Cmd.list with no options) being the right choice to run a group of Cmds. Occasionally, you may want to run them in series, and occasionally, you want to batch the results and send them at the end, but those should be an exceptional choice choice and not worth adding a whole new top level Cmd for. Having both in the top level API makes the seem equally normal choices when they are not. 
- The old Cmds both have batching enabled, but I'd like the new default to be batching disabled (and parallel not series). It's harder to change the default behavior if we don't deprecate the old methods because we can't warn in the previous major version of deprecation and force them to address the new behavior.

farism:
I think that @jarekb84 brings up a good point, but I believe that breaking changes are going to happen, and as long as we stick to semver, proper docs, and a deprecation cycle, it's not a huge deal. We could even backport the configurable batching for `sequence` and `batch` and default to true to give current `3.0` and `4.0` api consumers the feature.

But I mostly agree with @bdwain on all three points, and particularly the first one. Running a list of commands is a single concept, and the way and order they are resolved feels like a piece of configuration.

Deprecating the old API isn't just about simplifying it but also about training users on the generic concept.

ashishsc:
I think this is a fine change. Breaking changes will happen as @farism change. I personally prefer to have separate names for the functions rather than have an option object, but I agree that from a user's perspective it doesn't matter until they _need_ to dig deeper

bdwain:
Cool. Thanks for the feedback everyone. I think I'm going to go with the single list Cmd like originally planned. It's a little more verbose if you don't want to un the Cmds in parallel, but I think that's a good thing because it'll encourage people to use the parallel non-batched version by default, . Almost all of the the time, you should be able to use Cmd.list with no options.

index.d.ts
willisplummer: 👍 makes sense to me

Issue #137, opened 2017-08-04
Improve testability of Cmd objects


jon-thompson:
While it's comparatively easier to test reducers with redux-loop than with some other pieces of middleware, I do think it could be improved. Adding and removing specs for `Cmd` objects can sometimes be a little clunky, and some things are just not possible to test without inspecting the (essentially private) members of the `Cmd` object.

Some problems I run into when testing:

### 1. Action creators that take arguments

In some cases, I want a `Cmd` to dispatch an action with a predetermined argument. As it stands right now, I'd try to write a spec that looks like this:

```
expect(getCmd(reducer(state, action))).toEqual(Cmd.run(submitForm, {
  args: [Cmd.getState],
  successActionCreator: () => formSubmitted('FORM_NAME'),
}));
```

However, the above spec will always fail because `successActionCreator` is always a new function so those properties will never match. It would probably be nice to have a way for the spec to directly access `successActionCreator` so it can execute it and assert against the result. 

Possible solutions:

* A property/function on the `Cmd` object that returns the action creators
* A helper function like `getCmd` that returns the action creators

### 2. Multiple commands in the same loop

I often find myself adding and removing commands from a loop, so I prefer to have one test/assert per command. That is a little challenging with `redux-loop` as it is now though, because we only have the `Cmd` object to work with so we have to assert all commands at the same time. I currently use the `cmds` property on the `Cmd.batch` object but I don't think that is truly supported and is less than ideal because `Cmd.run` does not have a `cmds` property.

Possible solutions:

* A supported property/function on the `Cmd` object that returns an array of the commands (regardless of the object's type)
* A helper function like `getCmd` that returns all the commands, regardless of how they were grouped

I'd love to be able to do something like `expect(getCmds(reducer(state, action))).toContain(Cmd.run(someFunction))` where `getCmds` always returns an array whether the loop had a single `Cmd.run` or `Cmd.batch` or `Cmd.sequence` or just returned model state. Collections for the win! :) 

---

While these problems are kind of separate, I opened a single issue for them because I think the solution for one may inform the solution of the other. Overall, it'd be great to push forward a more concrete, supported testing strategy for `Cmd` objects, but feel free to split this up into separate issues as you see fit.

bdwain:
[...]

PR #152, opened 2017-09-30
Result action creator with extra args

gniquil:
Hi All,

Here's a pull request to add extra args to success/fail action creators. Briefly, the following example from the API doc should show the feature:

```js
import { loop, Cmd } from 'redux-loop';

function fetchUser(userId) {
  return fetch(`/api/users/${userId}`);
}

function userFetchSuccessfulAction(user) {
  return {
    type: 'USER_FETCH_SUCCESSFUL',
    user
  };
}

function userFetchFailedAction(err, { userId }) {
  return {
    type: 'USER_FETCH_ERROR',
    err,
    userId
  };
}

function reducer(state , action) {
  switch(action.type) {
  case 'INIT': {
    const userId = '123';

    return loop(
      {...state, initStarted: true, userId},
      Cmd.run(fetchUser, {
        successActionCreator: userFetchSuccessfulAction,
        failActionCreator: [userFetchFailedAction, { userId }],
        args: [ userId ]
      })
    );
  }

  case 'USER_FETCH_SUCCESSFUL':
    if (action.user.id !== state.userId) {
      // ignore fetch result if it's for a different user
      return state;
    }
    return {...state, user: action.user};

  case 'USER_FETCH_FAILED':
    if (action.userId !== state.userId) {
      // ignore fetch error if it's for a different user
      return state;
    }
    return {...state, error: action.error};

  default:
    return state;
  }
}
```
Note that now `failActionCreator` is now an array of 2 elements first being the result action creator, and second element being the extra argument (works exactly the same for `successActionCreator`. Also, you can take a look at the typescript definition to see how it helps you type check (as much as I can manage to make it work).

Frank


bdwain:
I'm not sure we should be complicating the API used in code just to to simplify testing. I think it's simpler more intuitive to pass something like `result => ac(result, foo)` to a Cmd than `[ac, foo]`.

Checking function equality in tests is always difficult. I think it'd be better to go with the approach we discussed in #137 where we can simulate a cmd to verify the action creators on it. It wouldn't change the regular API (only for testing), and we would still be able to check for equality in cases where we don't have nested action creators, but could use the simulations for those more complicated scenarios. Something like

```js
  let result = reducer(state, action);
  const expectedLoop = loop(newState, Cmd.run(func, {
    successActionCreator: jasmine.any(Function)
  }));
  expect(result).toEqual(expectedLoop);
  expect(getCmd(result).simulate({success: true, result: 123})).toEqual(ac(123, hardCodedParam));
```

I apologize for not responding to your issue sooner. I do appreciate the PR though. 

Addressing the issues in #137 is the top priority for me right now so hopefully we can address them soon. If you're interested in implementing any of the solutions from there, feel free to open a PR and I will review it ASAP.

I'm going to close this PR for now. If you want to continue to discuss this approach or other approaches to improving testability, please continue the conversation in #137. We can reopen this PR later if necessary, but I'd like to keep the discussion in a single thread.

PR #154, opened 2017-10-01
Improve testing

bdwain:
Addresses some of the issues in #137. Enables simulation of cmds and disabling invariants for testing matchers.

 @gniquil and @jon-thompson I haven't done the tests or doc changes yet. I want to see what you think of this implementation first.

Basically, now you can say

```js
//Cmd.run
expect(getCmd(result).simulate({result: 123, success: true})).toEqual(actionCreator(123));
expect(getCmd(result).simulate({result: 123, success: false})).toBe(null);

//Cmd.actiom
expect(getCmd(result).simulate()).toEqual(actionIPassedToCmd);

//Cmd.list
expect(getCmd(result).simulate([
  {result: 123, success: true},
  {result: 456, success: false}
])).toEqual([actionCreator1(123), actionCreator2(456]);
```

alternatively, we can expose something like Cmd.simulate(cmd, simulation), but I kind of like it on the cmd itself. Another option is to break this logic out into a separate package, but I think it's simple enough to not bother.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,40 @@ state and always get a result which is `deepEqual`.
 > not equal within JavaScript, and so are best to avoid if you want to compare
 > effects in your tests.
 
+#### Simulating Cmds
+
+Occasionally you may find yourself in a situation where you need to pass more than one parameter to an action creator, such as:
+
+```js
+Cmd.run(foo, {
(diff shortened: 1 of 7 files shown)


gniquil:
Works for me too.

PR #155, opened 2017-10-02
fix RunCmd and ListCmd types

willisplummer:
quick fix to the types of interfaces returned by ListCmd and RunCmd -- shouldn't affect many end users because it's mainly just handled internally by `redux-loop`, but `options` gets spread so `args` wouldn't exist inside of an options object here for example:

```
  return Object.freeze({
    [isCmdSymbol]: true,
    type: cmdTypes.RUN,
    func,
    ...options
  })
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -25,10 +25,8 @@ export interface NoneCmd {
 export interface ListCmd<A extends Action> {
   type: 'LIST';
   cmds: CmdType<A>[];
-  options: {
-    sequence: boolean;
-    batch: boolean;
-  };
+  sequence?: boolean;
+  batch?: boolean;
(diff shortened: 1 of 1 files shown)


bdwain:
good catch!

bdwain:
@willisplummer forgot to mention this before, but in the future, can you use the rebase and merge option rather than create a merge commit? i feel like merge commits just make the history harder to scroll through.

doesn't matter for this time, but in general I'd rather have that be the norm.

willisplummer:
👍 @bdwain could also enable to squash merge feature?

bdwain:
sweet i didn't know about that option

Issue #158, opened 2017-10-05
Replace readme with a multi-page guide 


bdwain:
The top-level readme is getting bloated trying to explain every feature. It'd be great to have a more comprehensive guide and leave the readme to be a high level explanation and some links.

A gitbook would be cool, but maybe unnecessary considering redux-loop is a relatively small library. 
I think long term, we'll want one though.

At the very least the docs folder could be expanded upon to have pages that are focused on specific topics or concepts.

bdwain:
i set up gitbooks for the project. I'm gonna play around with that for a bit and get a js.org domain to host it on.

on3iro:
Sounds great. I was actually thinking that we should spread the word about this awesome library a bit more. So maybe creating some kind of logo would be cool as well. Unfortunately I am not a designer so, I can't really do that, but I could ask some friends.

If you need support with the docs, let me know :)

bdwain:
hm there is a logo and i thought it was in the readme. I guess not. We should add it back. 

And yes I've been meaning to work on the docs but haven't had the time. If you're interested in making a PR to add some docs that would be very much appreciated (if so, don't worry about the gitbook for now. Markdown files are fine and can be converted to a gitbook easily).

on3iro:
@bdwain sounds good to me. I will see if I can find some spare time over the next week to work on the docs :)

psbanka:
I'm also interested in working on this and I think I have some spare time to contribute in the next few days, @on3iro – hit me up if you'd like to work on this together?

on3iro:
@psbanka cool. Looks like you already put some work into this :) Unfortunately I am on vacation the next two days, but could help out on saturday. Just hit me up if there is anything I can help you with.

bdwain:
This is complete. Thanks again @psbanka !

Issue #170, opened 2017-12-11
Built working example with create-react-app, redux-loop and TypeScript


meandmax:
May be interesting for others, had some barriers to setting this up: https://github.com/meandmax/redux-loop-ts-example

bdwain:
sweet! I'll link to this from the readme.

bdwain:
Actually I'll leave this open for now because we there's not really a good spot for it in the current readme, which needs to be broken up into better docs. Once that's done, i'll add this link to that.

bdwain:
This is linked to now on https://redux-loop.js.org/docs/tutorial/Tutorial.html

We can probably do a better job of highlighting typescript support in general, but I'll leave that to separate issues about improving the organization and clarity of the docs overall. Thanks again for creating that example @meandmax !

PR #171, opened 2017-12-26
Issue 158: better docs

psbanka:
Related to https://github.com/redux-loop/redux-loop/issues/158

1. Adding build targets in package.json for gitbook
2. Set up basic gitbook infrastructure, cribbed primarily from redux
3. Broke README.md up into several files in docs/tutorial
4. Added favicon.ico, reduced down from main logo
5. Added dependencies for gitbook and related plugins 
  - github (for adding a link to the repos), 
  - anchor (for adding links-on-hover to sections)
  - edit-link (for giving people a shortcut to the original `.md` file in the source for easy update PRs)
  - prism (for better code highlighting)
6. Added base css for gitbook
7. Added table-of-contents in SUMMARY.md
8. Updated messages in `src/cmd.js` for new locations of docs


bdwain:
Sorry for the delay @psbanka . I was pretty busy over the holidays and am just now getting a chance to look at this.

At first glance it looks great! Thanks so much for your efforts getting the docs set up.

I haven't had the chance to look at everything in detail yet. I'll try to finish doing that over the weekend. I have a few small initial comments though.

docs/api-docs/cmds.md
```diff
@@ -1,206 +1,19 @@
...
 
 The `Cmd` object provides access to all of the functions you'll need to
 represent different kinds of cmds to redux-loop's cmd processor. Every
 cmd is a plain JavaScript object that simply describes to the store how to
 process it. Cmd are never executed in the reducer, leaving your reducer pure
 and testable.
 
+## Command objects
+
+These are things that can be loop'ed
```
bdwain: This part is a little confusing. Is this describing the result of Cmd.foo vs Cmd itself?

docs/api-docs/combineReducers.md
```diff
@@ -0,0 +1,27 @@
...
+
+```js
+import { combineReducers } from 'redux-loop';
+import reducerWithSideEffects from './reducer-with-side-effects';
+import plainReducer from './plain-reducer';
+
+export default combineReducers({
+  withEffects: reducerWithSideEffects,
+  plain: plainReducer
+});
```
bdwain: I think it'd be good to mention the immutable/other helper libraries here too.

docs/api-docs/combineReducers.md
```diff
@@ -0,0 +1,27 @@
...
+* `reducersMap: Object<string, ReducerFunction>` &ndash; a map of keys to nested
+  reducers, just like the `combineReducers` you would find in Redux itself.
+
+## Notes
+
+Reducer composition is key to a clean Redux application. The built-in Redux
+`combineReducers` won't work for nested reducers that use `loop`, so we included
+one that is aware that some reducers might have side effects. The `combineReducers`
+in redux-loop knows how to compose cmds as well as state from nested reducers
+so that your effects tree is always separate from your state tree. It's also
```
bdwain: I've always felt that this line is confusing. 

It's also completely compatible with the one in Redux, so there should be no issue switching to this implementation.

I think we should reword it or get rid of it. I think it was intended to point out that if something out of your control is using vanilla combineReducers and has no loops inside of it, then you can pass the result of that to the redux loop version. But the wording just seems confusing to me.

docs/tutorial/AdvancedTopics.md
```diff
@@ -0,0 +1,117 @@
...
+
+This minimal example will cause perpetual dispatching! While it is also possible
+to make this mistake with large, complicated networks of `redux-thunk` action
+creators, it is much easier to spot the mistake before it is made. It helps to
+keep your reducers small and focused, and use `combineReducers` or manually
+compose reducers so that the number of actions you deal with at one time is
+small. A small set of actions which initiate a `loop` will help reduce the
+likelihood of causing circular dispatches.
+
+## combineReducers with redux-loop
```
bdwain: I think combineReducers should be in the tutorial and not advanced topics. It's basically required unless you make heavy use of Cmd.map or do something else creative, which someone using the tutorial probably wouldn't don

I've also had a bunch of people make issues about things not working and the problem was using vanilla combine reducers, so I think it'd be good for us to be more clear on that.

docs/tutorial/AdvancedTopics.md
```diff
@@ -0,0 +1,117 @@
...
+
+`combineReducers` just returns a reducer with a specific usage of `mergeChildReducers` (one in which
+the child map never changes, the parent state is an empty object, and all children always get called). 
+By using `mergeChildReducers` in your reducer, you can choose which children are used when and under what key names.
+
+By using `combineReducers` or `mergeChildReducers` everywhere your reducer tree nests, all Cmds resulting 
+from a single dispatched action will always be batched up into a single Cmd at the top of the tree and 
+processed correctly by the store. 
+
+## Using `combineReducers` and `mergeChildReducers` with ImmutableJS and other libraries
```
bdwain: What ar eyour thoguhts on moving the majority of the detail here to the api page about combine reducers and then making a small mention of it on the tutorial page instead of here (and a link to the detailed page). I think so many people will use immutable that it's worth adding to the tutorial so they don't have to go looking for it.

docs/tutorial/README.md
```diff
@@ -0,0 +1,120 @@
...
+  case 'USER_FETCH_FAILED':
+    return {...state, error: action.error};
+
+  default:
+    return state;
+  }
+}
+```
+
+Any reducer case can return a `loop` instead of a state object. A `loop` joins
```
bdwain: i would remove the word case here.

bdwain:
i'm not sure the best place for it but i'd like to add a link to https://github.com/meandmax/redux-loop-ts-example (from #170).

bdwain:
can you delete docs/api-docs/README.md ? it's empty

docs/api-docs/cmds.md
```diff
@@ -1,206 +1,19 @@
...
- top of redux-loop's API, or for just writing your own combineReducers.
-
-
-
-## `Cmds`
-
-#### Notes
+# `Cmds`
 
 The `Cmd` object provides access to all of the functions you'll need to
```
bdwain: I know you didn't write this part but I think the wording is a bit confusing. How about

The `Cmd` object provides access to all of the helpers you need to create cmd. A cmd is a plain object describing a side effect that can happen after your reducer runs. It is executed by the enhanced redux store outside of the reducer, which leaves the reducer pure and testable.

docs/api-docs/cmds.md
bdwain: if you make the change above (rewording the first paragraph) I think you might be able to just remove this part about Command objects

docs/api-docs/cmds.md
```diff
@@ -261,16 +75,20 @@ return loop(
 );
 ```
 
-
 ### `Cmd.run(func, options)`
```
bdwain: I just tried to read through the Cmd documentation and I feel like it would be easier to read if we moved the notes for each one above the parameters and the simulation instructions. Sorry I know you didn't do that either. It was just weird having to read a bunch of usage details before seeing an explanation of what it does.

docs/api-docs/combineReducers.md
bdwain: or do you think it would read better to have a short description above and the detailed notes below (where they are now)

docs/api-docs/cmds.md
bdwain: or do you think it would read better to have a short description above and the detailed notes below (where they are now)

docs/api-docs/getCmd.md
```diff
@@ -0,0 +1,11 @@
+# `getCmd(loop): Cmd | null`
+
+* `loop: any` &ndash; any object.
+* returns the cmd component of the array if the input is a `[any, cCmd]`
```
bdwain: cCmd is a typo I think. I would maybe say 

if the input is a `loop` object (`[any, Cmd]` pair)

docs/api-docs/liftState.md
```diff
@@ -0,0 +1,43 @@
+# `liftState(state): [any, Cmd]`
+
+* `state: any` &ndash; an object which may be the state of the redux store, or
+  an existing `[any, Cmd]` pair created by `loop()`.
+
+## Notes
+
+Automatically converts objects to `loop()` results. If the value was created
+with `loop()`, then the function behaves as an identity. Otherwise, it is lifted
+into a `[any, Cmd]` pair where the effect is `Cmd.none()`. Useful for
```
bdwain: can you remove the parentheses on Cmd.none? It's not a function

docs/api-docs/liftState.md
```diff
@@ -0,0 +1,43 @@
...
+      };
+    default:
+      return state;
+  }
+}
+
+// This guarantees that the return value of the reducer will be a loop result,
+// regardless of if it was set as such in the reducer implementation. This makes
+// it much easier to manually compose reducers without cluttering reducer
+// implementations with `loop(state, Cmd.none())`.
```
bdwain: here too with the parentheses on Cmd.none

SUMMARY.md
```diff
@@ -0,0 +1,24 @@
+# Summary
```
bdwain: I don't see any links to this file. Will it be used by the gitbook?

psbanka:
Wow, great feedback, @bdwain – I'll try to address everything you brought up.

docs/api-docs/combineReducers.md
psbanka: I'll take a stab at this, but I didn't use the immutable libraries myself, so I'm not completely certain how they're used.

docs/tutorial/AdvancedTopics.md
psbanka: Agreed!

docs/tutorial/AdvancedTopics.md
psbanka: yes, I think that's smart. Will update.

docs/api-docs/cmds.md
psbanka: I think this can be confusing because we have objects that are generated by functions like `Cmd.none()`, `Cmd.action()`, `Cmd.list()`, etc. It seems to me that those object which are generated by those functions are really what we should refer to as "command objects". We have a _module_ called `Cmd` which provides a bunch of convenience functions for generating command objects. Therefore I think it might make more sense to say something more like this:

> The `Cmd` module provides access to all of the functions you'll need to
create and represent different kinds of cmd objects to redux-loop's
cmd-processor. Every cmd object is simply a plain JavaScript object that describes a side-effect that can happen after your reducer runs. This cmd object is then executed by the enhanced redux store outside of the reducer, which leaves the reducer pure and testable.

docs/api-docs/cmds.md
psbanka: yeah, no, I agree with your first idea. Let me give that a shot first and see how it reads..

docs/api-docs/liftState.md
psbanka: Oh, right! will do.

SUMMARY.md
psbanka: Yes, if you do a `npm run build:watch`, you can see the `SUMMARY.md` file in use in the left-hand side:

screen-shot

docs/api-docs/cmds.md
bdwain: sounds good to me!

package.json
```diff
@@ -83,7 +88,13 @@
     "typescript": "^2.5.3",
     "typescript-definition-tester": "^0.0.5"
   },
-  "dependencies": {},
+  "dependencies": {
+    "gitbook-cli": "^2.3.2",
```
bdwain: can you move these to dev-dependencies? otherwise they'll get installed by everyone who uses redux-loop

package.json
```diff
@@ -30,7 +30,12 @@
     "build:umd:min": "cross-env BABEL_ENV=es NODE_ENV=production rollup -c -i src/index.js -o dist/redux-loop.min.js",
     "build": "yarn build:commonjs && yarn build:es && yarn build:umd && yarn build:umd:min",
     "prebuild": "yarn run clear",
-    "prepublishOnly": "yarn lint && yarn test && yarn build"
+    "prepublishOnly": "yarn lint && yarn test && yarn build",
+    "docs:clean": "rimraf _book",
+    "docs:prepare": "gitbook install",
+    "docs:build": "npm run docs:prepare && gitbook build -g redux-loo/redux-loop && cp logo/logo.png _book/gitbook/images/logo.png && cp logo/favicon.ico _book/gitbook/images",
```
bdwain: since redux-loop uses yarn, can you switch `npm run docs:prepare` to `yarn docs:prepare`  and all of the other npm run's to yarn too? Same with the other tasks.

also, looks like a type where it says `redux-loo/redux-loop` instead of `redux-loop/redux-loop` in a few spots.

package.json
```diff
@@ -30,7 +30,12 @@
...
+    "docs:watch": "npm run docs:prepare && gitbook serve",
+    "docs:publish": "npm run docs:clean && npm run docs:build && cp CNAME _book && cd _book && git init && git commit --allow-empty -m 'update book' && git checkout -b gh-pages && touch .nojekyll && git add . && git commit -am 'update book' && git push git@github.com:redux-loo/redux-loop gh-pages --force"
```
bdwain: can you add a file to the root of the repo called CNAME that contains `redux-loop.js.org`? This task is looking for that file. After this is working I'll get that url from https://dns.js.org/

bdwain:
can you delete package-lock.json? Must have been created because the redux tasks you copied used npm not yarn. I'll add a gitignore for the future.

bdwain:
i think the links got messed up in the summary file. The api-docs link currently points to a file that does not exist, and that is causing the whole section to be disabled.

screen shot 2018-01-29 at 11 50 03 pm

Also, the tutorial section link points to the same file as the writing the reducer link.

I think it'd be good to follow redux's pattern there of having a README in each folder that the section link points to, which just repeats the sub-links and maybe has a short blurb about the section (i don't think either of our sections need the blurb). And adding a readme will also fix the disabled links.

For the tutorials section, maybe we can just rename it getting started, and point the getting started link at a readme section.

SUMMARY.md
bdwain: looks like redux made SUMMARY.md a symlink to docs/README.md. I kind of like that since it keeps all of the docs in one folder. Do you mind doing that too?

docs/api-docs/combineReducers.md
bdwain: sorry I should have given more detail. Maybe change it to something like this.

### Using Immutable.js

If you're using
[immutable.js](https://github.com/facebook/immutable-js/) with your reducer,
you'll want to use the ImmutableJS version of combineReducers from the [redux-loop-immutable](https://github.com/redux-loop/redux-loop-immutable/) package. It has the same API as the non-immutable version except that the state it returns is an Immutable `Map` instead of an `Object`.

```js
import { combineReducers } from 'redux-loop-immutable';
import reducerWithSideEffects from './reducer-with-side-effects';
import plainReducer from './plain-reducer';

export default combineReducers({
  withEffects: reducerWithSideEffects,
  plain: plainReducer
});
```

bdwain:
This is looking great @psbanka ! I think after my last batch of comments we should be good to go. Once this is out it'll be much easier for us to improve the content and quality of the docs. Thanks again for getting this started. You've been a huge help and I really appreciate it.

psbanka:
Okay, @bdwain – have a look. I think I caught all your changes. lmk if there's anything missing :)

bdwain:
Sweet! @psbanka. I'll take a look tonight.

bdwain:
looks good to me! I added 2 small commits with some minor touch ups but that's it. thanks so much @psbanka ! This has been super helpful!

bdwain:
redux-loop.js.org should be live soon! https://github.com/js-org/dns.js.org/pull/1745

PR #177, opened 2018-03-01
Add clear cmdsQueue after replace reducer

Fyzu:
After dynamically replace reducer and dispatch any action, old cmds running again.

Example: http://s.csssr.ru/U5E1S7FNC/20180301163042.png
In example, after "LOCATION_CHANGE", i will replace reducer and call dispatch.
Old cmd ''[3] items/ACTION", running again.

### DIFF
--- a/src/install.js
+++ b/src/install.js
@@ -51,7 +51,8 @@ export function install() {
     }
 
     const replaceReducer = (reducer) => {
-      return store.replaceReducer(liftReducer(reducer))
+      store.replaceReducer(liftReducer(reducer))
+      cmdsQueue = []
     }
 
     runCmd({


bdwain:
Are you using redux-devtools? That is typically the cause of the cmds re-running after replacing the reducer. It's because redux devtools re-dispatches every action when you replace the reducer. You can avoid that by setting shouldHotReload to false. It's an option that can be used for the chrome extension or for regular redux devtools.

Fyzu:
Yes, i used redux-devtools.
But it ok when re-run only cmd result actions?

bdwain:
it doesn't re-run cmds. redux-devtools creates its own instance of your store. so if you re-dispatch all actions, some of those will trigger cmds, which will cause side effects to get re-run. turning off shouldHotReload would solve the problem, it will

Fyzu:
But, why does this happen only after replaceReducer and call dispatch any action?

bdwain:
because when you call replaceReducer, redux-devtools re-dispatches every action it ever saw. so if you handle action FOO by running a Cmd, when you call replace-reducer, redux-devtools redispatches FOO, which reruns the Cmd. If you turn off shouldHotReload, it will prevent the devtools from redispatching FOO, so the Cmd won't run again.

skoestler:
For what its worth, I tried the suggested solution of setting shouldHotReload to false and it has the unfortunate effect of clearing the the inspector log after replacing the reducer.  Making this change definitely hurts the utility of the dev tools.

bdwain:
yea I've noticed that. I filed an issue a long time ago, but it hasn't been addressed. https://github.com/reduxjs/redux-devtools/issues/378

If the action you want to inspect is getting erased, i recommend adding https://github.com/LogRocket/redux-logger to your app in dev mode

PR #179, opened 2018-03-17
Pass whole state as 3rd param for reducer in combineReducers

moljac024:
This lets sub-reducers access the whole state. It is useful in some
situation where you want to separate conceptual models but you still
need some bit of info from a sibling reducer.

For example, I like to have a separate auth reducer where I handle user specific actions and where I store the logged in user etc.
But any given reducer might need information about logged in user for some of its updates and it can be much more code to thread the user from the ui and pass it into actions etc.
In fact, this is the exact use-case that I have right now and where I feel this third parameter would be useful.

It would let me write:

```javascript
export const reducer = (state = initialState, action, wholeState) => {
  switch (action.type) {
    case types.SEND_MESSAGE:
      const id = uuid.v4()
      return loop(
        {
          ...state,
          text: '',
          messages: [
            ...state.messages,
            {
              message_id: id,
              showAvatar: true,
              from_id: wholeState.auth.user.id,
              message: state.text,
              sending: true,
            }
          ]
        },
        Cmd.none // Actual Cmd is not interesting here
      )
    default:
      return state
  }
}
```

Unless I'm mistaken, If I didn't have the whole state here as 3rd parameter I would need to make up a second action to which I would pass Cmd.getState as a parameter, have that eventually dispatch this with the user id as argument and that sounds way too convoluted to be useful and clean enough.

### DIFF
--- a/src/combineReducers.js
+++ b/src/combineReducers.js
@@ -42,7 +42,7 @@ export const combineReducers = (
     const model = Object.keys(reducerMap).reduce((model, key) => {
       const reducer = reducerMap[key]
       const previousStateForKey = accessor(state, key)
-      let nextStateForKey = reducer(previousStateForKey, action)
+      let nextStateForKey = reducer(previousStateForKey, action, state)
 
       if (isLoop(nextStateForKey)) {
         cmds.push(getCmd(nextStateForKey))


bdwain:
In general I'd rather not deviate from the behavior of redux's combineReducers when possible, especially because this problem isn't specific to redux-loop. There are different ways to solve this problem, but I don't think redux-loop should be choosing one. Especially because this would change the typings for reducers used by redux-loop to need a third parameter that most people would not use. 

[Cmd.getState](https://redux-loop.js.org/docs/api-docs/cmds.html#cmdgetstate) and [Cmd.dispatch](https://redux-loop.js.org/docs/api-docs/cmds.html#cmddispatch) are escape hatches that you can use to access arbitrary slices of state and dispatch arbitrary actions from a side effect function. At the very least, you can use that, though in general I'd say that's not ideal.

Personally, I wrote a small dependency-injection middleware called [redux action enhancers](https://github.com/bdwain/redux-action-enhancer) and have been using that for a while to solve problems like this. It's pretty nice for adding common pieces of state to an action without explicitly passing it in. Check it out if you're interested. 

If none of that works, I would recommend writing your own middleware or own version of combineReducers to do exactly what you want it to. CombineReducers is just a generic helper that's meant to solve the common case. It's certainly not right for everything or everyone, and it's not uncommon to need a custom one that does something different.

bdwain:
I'm going to close this as I don't think any change is necessary. But we can continue to discuss here if you'd like.

PR #182, opened 2018-04-13
Made Cmd types more precise

NotWoods:
- Marked `CmdType` properties as readonly, since a frozen object is returned
- Used unique `simulate()` functions on each `CmdType` interface.
- `Cmd` is now a namespace instead of a static class, to reflect that its exported as an object.
- Fixed typescript tests

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -79,7 +79,7 @@ export type CmdType<A extends Action> =
   | RunCmd<A>
   | BatchCmd<A>
   | SequenceCmd<A>;
-  
+
 declare function install<S>(): StoreEnhancer<S>;
 
 declare function loop<S, A extends Action>(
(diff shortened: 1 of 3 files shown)


bdwain:
Sorry i haven't been able to look. I've had a busy few weeks. I'll try to get to it this weekend. Thanks!

bdwain:
This all seems to make sense. Is any of this a breaking change though? Specifically the second bullet. 

@ashishsc or @willisplummer thoughts?

NotWoods:
The simulate functions won't be breaking unless they were used improperly. Here's a breakdown of how they change for each Cmd type:

### `NoneCmd.simulate`
```typescript
// Before
(simulation?: CmdSimulation) => AnyAction
// After
() => null
```
None's simulate is meant to be used without a parameter, so if the user was accidentally supplying a parameter then typescript will now catch that error. If they used the return value of simulate and they are using strict null checks, then it will now properly indicate that the type is null. 

### `ListCmd.simulate`
```typescript
// Before
(simulations?: MultiCmdSimulation) => A[]
// After
(simulations: MultiCmdSimulation) => A[]
```
This now enforces that a parameter is given to the simulation function. Again, this is the intended use but previously the type checker allowed you to omit the parameter. Return type is identical.

### `ActionCmd.simulate`
```typescript
// Before
(simulation?: CmdSimulation) => A
// After
() => A
```
ActionCmd's simulate is meant to be used without a parameter, so if the user was accidentally supplying a parameter then typescript will now catch that error. Return type is identical.

### `MapCmd.simulate`
```typescript
// Before
(simulation?: CmdSimulation) => A | (simulations?: MultiCmdSimulation) => A[]
// equivalent to (simulations?: CmdSimulation | MultiCmdSimulation) => A | A[]
// After
(simulations: CmdSimulation | MultiCmdSimulation) => A[] | A | null
```
This now enforces that a parameter is given to the simulation function. This is the intended use but previously the type checker allowed you to omit the parameter. If they used the return value of simulate and they are using strict null checks, then it will now properly indicate that the type can null. 

### `RunCmd.simulate`
```typescript
// Before
(simulation?: CmdSimulation) => A
// After
(simulation: CmdSimulation) => A
```
This now enforces that a parameter is given to the simulation function. Return type is identical.


### TL;DR
For parameters: usages that would throw a runtime error now also have a type error.  
For return types: functions that return null now properly indicate it. This will affect code that assumes the return type is always not null.

bdwain:
Makes sense! Thanks for the detailed explanation. Let me look at it in the next day or two but then we should be good to go.

Issue #181, opened 2018-03-29
Make a recipies section in the docs


bdwain:
We should have a recipes section with common patterns/problems and solutions to them.

Some ideas to start

#176 request queue
#180 polling

on3iro:
I am currently quite busy, but confident that I'll be able to look into this by the end of next week :)

bdwain:
Closed via #183

Issue #176, opened 2018-02-14
[Question]: How would I implement a request queue


on3iro:
Hey, 

I am currently having a rather specific problem that I am not quite sure of how I should tackle it.
I am building some kind of task board and whenever a task is dragged around, I will immediately update the store and run a command to initiate a server request. If the request is unsuccessful the old state will be restored.
However because of the asynchrounous nature of these sideeffects it might be, that two tasks fail in quick succession, and the second tasks old-state will overwrite the first tasks old state, which leads to an overall wrong application state. 

Now I am trying to think of a way to queue my sideffects. So that these specific requests do not run simultaneously and are instead handled one after another.
Is there a recommended way of how to approach a problem like this with redux-loop?

Thanks in advance!

bdwain:
Hi @on3iro . I'm assuming that these side effects are triggered independently. One event does not trigger 2 side effects, right? Because in that case, you could just use [Cmd.list](https://redux-loop.js.org/docs/api-docs/cmds.html#cmdlistcmds-options) with the sequence option.

Assuming that the side effects are triggered independently, I think you actually said the answer already. You want to queue them up, which is something that can be represented in your application's state. 

For example, let's say you dispatch 2 events in quick succession, one saying that item 123 was dragged, and one saying item 456 was dragged. There would be an ITEM_MOVED action, a REQUEST_SUCCESS action, and a REQUEST_FAILED action to handle. 

On ITEM_MOVED, you'd either start a request or add it to the back of the queue, depending on if there are items in the queue already. On REQUEST_SUCCESS, you'd remove the old request off of the front of the queue and start the next one (if there was a next one). On REQUEST_FAILED, you'd throw away the rest of the queue (or you could continue on in a different way if you wanted to).

Here's how you could handle those 3 types of actions

```js
const initialState = {
   queue: [],
   error: false,
   //other fields
};

function handleItemMovedAction(state, action){
  let queue = state.queue;
  if(queue.length){ //something is already running so just add a request to the queue
    return {
      ...state,
      queue: queue.concat([action.id]) //add new request to back of queu
    };
  }
  //nothing is running so just add the new request to the front of the queue and run it
  let newState = {
    ...state,
    queue: [action.id]
  };
  return loop(newState, Cmd.run(doRequest, {
    args: [action.id],
    successActionCreator: requestSuccessAction,
    failActionCreator: requestFailAction
  }));
}

function handleRequestSuccessAction(state, action){
  let newState = {
    ...state,
[...]

PR #183, opened 2018-04-18
Add simple command queue example

on3iro:
Reference: #181 
@bdwain 
I just added the request queue example for now. I also tried to make it a bit more verbose in comparison to the answer you gave me in #176 .
Could you please have a look if the example is comprehensible and if you approve the coding style of the examples?
I will probably open another PR for the polling example after we are done with this one :)


bdwain:
Thanks! Sorry I haven't had a chance to look. I'll try to do it this weekend.

on3iro:
No worries - take your time :)

docs/README.md
```diff
@@ -5,6 +5,7 @@
   * [Tutorial](/docs/tutorial/Tutorial.md)
   * [Testing](/docs/tutorial/Testing.md)
   * [Advanced topics](/docs/tutorial/AdvancedTopics.md)
+  * [Recipes](/docs/tutorial/Recipes.md)
```
bdwain: I was thinking Recipes would be a top level section like API Docs with sub pages rather than a big page under Getting started. Thoughts?

bdwain:
So it looks like this example is queueing multiple types of requests, and not just the single type from the task board dragging. That's a bit more complex than a single queue item type, so I think it'd probably be worth starting with an example of that before showing how to generalize it to multiple types of actions. I think the example I gave in #176 would be good for the simple case (with maybe a bit of cleaning up). Do you want to add that on the top of the page? Or if it makes sense we could also make them 2 separate pages in the recipes section. 

There's also 2 issues with your example that I'd like to address. The first is that you're storing Cmds in state, which [should be avoided because they are not serializable](https://github.com/reactjs/redux/blob/master/docs/faq/Actions.md#actions-string-constants). Since Cmds usually have functions in them they should not be passed around and stored in state.

The other issue is that you need to dispatch 2 actions every time you want to trigger a queueable event, which is not ideal. It's more work for the user to remember, and generally you just don't want to dispatch multiple actions in response to the same event happening. It's better for an action to represent a thing that happened rather than have a thing that you want to happen. And then you would listen for the action in multiple spots rather than each action only listened to once. Then the action history becomes a sort of event log that's easy to see what things happened and what state changes they caused. It also helps make state updates atomic, which can reduce bugs and simplify debugging, (imagine if the LOGOUT action caused every other reducer to dispatch their own CLEAR_DATA action. they'd all have data in the state for a bit after you logged out).

I think both issues are kind of tied together, so we can probably fix them together too. Rather than store a list of Cmds, I think it would help to store a list of actions in a queue. The way you could accomplish that would be to mark actions to be queued separately from their type, so that both reducers could listen for them.

```js
export const attachQueuedAction = (action, queuedAction) => {
  return {
    ...action,
    queuedAction //the queue reducer looks for this instead of looking for type
  };
};

//abstract the queueing part of it away from the person dispatching this
//by having this action creator call the other one
export const itemDragged = id => attachQueuedAction({
   type: 'ITEM_DRAGGED',
   id
}, makeDragRequest(id));

const makeDragRequest = id => ({
   type: 'DRAG_REQUEST',
   id
});
```

and then the queue reducer could just store the underlying actions in a queue and dispatch them with Cmd.action when they are at the front of the queue. 

This would leave the issue of how you notify the queue that the request finished. You could also ignore the type here and instead add another field to the actions that indicated that they were a queueSuccess or a queueFailed action and the queue could listen for that rather than the type.

```js
const queueSuccess = action => ({...action, queueSuccess: true})
const queueFailed = action => ({...action, queueFailed: true})

const dragRequestSuccess = response => queueSuccess({
   type: 'DRAG_SUCCESS',
   response
});
//same for the failed one

//reducer
function reducer(state, action){
  if(action.queuedAction){
    //add to queue and potentially dispatch the action with Cmd.action if it was empty before
  }
   else if(action.queueSuccess){
     //remove from queue and potentially continue
  }
  else if(action.queueFail){
     //reset
  }
  return state;
}
```

one thing to be careful of there is to make sure that the fields on the addToQueue and queue success and queue failed actions are unique and don't accidentally get set on some random other action, though that could happen with type names too so i don't think it's too big of a concern.

let me know your thoughts on all of that!

on3iro:
Sorry for the late response, but I am currently quite busy (just started a new job ;D).

I just briefly skimmed over your suggestions, but they seem pretty reasonable to me. Dispatching 2 actions definitely felt a bit weird and uncomfortable, but I couldn't come up with a better solution.
Your solution looks much cleaner, though. Regarding the storage of CMDs: I completely overlooked the fact, that I store functions inside the state. Thanks for pointing that one out!

Thanks again for your response. I hope that I'll be able to take a closer look into this by the end of next week. I will split the whole thing into two examples like you suggested.

docs/README.md
bdwain: It's not really about importance. I just don't think specific recipes fit under getting started since you would most need them not just when getting started, but if you run into a problem later on. I  also think it would just be easier to navigate around that way.

bdwain:
Sounds good take your time and enjoy the new job.

on3iro:
Just a quick heads up: I haven't forgotten about this merge request. I am just super busy - may seems to be the month of appointments and family gatherings for me this year ;)
I really hope that I find some time to do this, soon.

bdwain:
no worries not a rush!

on3iro:
Hey @bdwain, I finally got the time to make the changes you suggested :) 👍 

Could you have another look, please?

docs/README.md
bdwain: @on3iro I think this is a dead link and needs to be removed

docs/tutorial/README.md
```diff
@@ -2,4 +2,5 @@
 
 * [Tutorial](Tutorial.md)
 * [Testing](Testing.md)
-* [Advanced Topics](AdvancedTopics.md)
\ No newline at end of file
+* [Advanced Topics](AdvancedTopics.md)
+* [Recipes](Recipes.md)
```
bdwain: @on3iro this is also a dead link and can be removed

docs/README.md
```diff
@@ -23,3 +24,5 @@
   * [`combineReducers`](/docs/api-docs/combine-reducers.md)
   * [`reduceReducers`](/docs/api-docs/reduce-reducers.md)
   * [`mergeChildReducers`](/docs/api-docs/merge-child-reducers.md)
+* [Recipes](/docs/recipes/README.md)
```
bdwain: can this go above API Docs?

docs/recipes/ActionQueue.md
```diff
@@ -0,0 +1,156 @@
...
+The example below shows, how you could model your initial state as well as your _QueueReducer_.
+
+```js
+const initialState = {
+  queue: [],
+  error: false,
+  //other fields
+};
+
+function handleItemDraggedAction(state, action){
```
bdwain: can you add the main reducer body (with the switch statement) up here? I think it'd be helpful to have the whole thing complete

docs/recipes/ActionQueue.md
```diff
@@ -0,0 +1,156 @@
...
+## Handling different kinds of actions
+
+Sometimes you might want to queue different kinds of actions.
+To achieve this, we could build a `QueueReducer` that listens for specific properties
+instead of an `action.type`:
+
+```js
+function QueueReducer (state, action) {
+  if (action.queuedAction) {
+    // add to queue and potentially dispatch the action with Cmd.action if the
```
bdwain: can you replace these comments with the actual logic? I think it might be easier to understand that way

docs/recipes/ActionQueue.md
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  return state
+}
+
+function SomeOtherReducer (state, action) {
+  if (action.type === 'ITEM_DRAGGED') {
+    // do something
+  } else if (action.type === 'DRAG_REQUEST') {
+    // return Cmd with successActionCreator and failActionCreator
```
bdwain: can you add the actual logic here as well?

docs/recipes/ActionQueue.md
```diff
@@ -0,0 +1,156 @@
...
+  } else if (action.queueFail) {
+    // reset queue
+  }
+
+  return state
+}
+
+function SomeOtherReducer (state, action) {
+  if (action.type === 'ITEM_DRAGGED') {
+    // do something
```
bdwain: maybe say something like `return updateDraggedState(state, action)` here

on3iro:
@bdwain no worries - I currently quite busy as well. I will incorporate the changes you considered asap (probably tomorrow or at the upcoming weekend) 👍

on3iro:
@bdwain I have implemented the changes. Could you have another look, please? I have not yet rebased my two commits, because I first wanted to make sure that my changes match your expectations :)

docs/README.md
```diff
@@ -5,6 +5,9 @@
   * [Tutorial](/docs/tutorial/Tutorial.md)
   * [Testing](/docs/tutorial/Testing.md)
   * [Advanced topics](/docs/tutorial/AdvancedTopics.md)
+  * [Recipes](/docs/tutorial/Recipes.md)
```
bdwain: @on3iro you missed this link. It's not pointing to a file so it's grayed out. It can just be removed

docs/recipes/ActionQueue.md
```diff
@@ -0,0 +1,203 @@
+# Action Queue
+
+## Building a simple action queue
+
+Sometimes it might be desirable to dynamically add commands to a queue and
+handle each one in succession. Lets say for example we want to be able to
```
bdwain: Let's not Lets

docs/recipes/ActionQueue.md
```diff
@@ -0,0 +1,203 @@
...
+drag tasks on a taskboard. Whenevever a drag movement ends we have to send the
```
bdwain: Whenever has a typo

bdwain:
@on3iro had a 3 last tiny comments and then I think this is good to go! Thanks!

on3iro:
@bdwain thanks for the quick review. I fixed the missing link and the typos and also squashed my commits.

bdwain:
Thanks @on3iro !

PR #189, opened 2018-07-27
Flow type signatures

stereobooster:
Not sure about correctness, haven't tried it out yet


index.flow.js
```diff
@@ -0,0 +1,149 @@
...
+  declare export type CmdType<A> =
+    | ActionCmd<A>
+    | ListCmd<A>
+    | MapCmd<A>
+    | NoneCmd
+    | RunCmd<A>
+    | BatchCmd<A>
+    | SequenceCmd<A>;
+
+  declare function install<S>(): StoreEnhancer<S>;
```
stereobooster: add export

flow-typed/npm/redux_v4.x.x.js
```diff
@@ -0,0 +1,59 @@
+// flow-typed signature: df80bdd535bfed9cf3223e077f3b4543
+// flow-typed version: c4c8963c9c/redux_v4.x.x/flow_>=v0.55.x
+
+declare module 'redux' {
```
bdwain: is it not possible to add these as an npm dependency? I don't think they should be part of redux-loop's src.

bdwain:
Thanks for creating this PR @stereobooster ! It'd be great for redux-loop to have flow type definitions. 

Do you know of any tools we can use to test these? I think it would be good for there to be automated tests around this the same way we have them for typescript definitions with the typescript-defintion-tester

flow-typed/npm/redux_v4.x.x.js
on3iro: There is a [central repository](https://github.com/flow-typed/flow-typed) where flow-types usually reside.

stereobooster: Some points:

- I'm afraid this is not possible to add as npm. This is how Flow signatures work right now, either inline (it is get downloaded from flow-typed) or provided by the package
- this is not essential here, can be removed. I added it for testing
- even if it will be part of this repository, it will be more like developer dependency it will not be picked up by end-user's Flow

stereobooster:
> Do you know of any tools we can use to test these?

I don't know of any 😞. The only thing I can think of is to wright some example Flow files and run checks against it using those type signatures. I will take a look what is possible

stereobooster:
To sum up:
- tests are done as sample Flow files, which run checks against provided signatures
- redux signatures are part of "test suite"
- added build script to copy signatures near the entry points
- I'm pretty happy with types except `combineReducers`, but I'll leave it for the next time

bdwain:
Thanks @stereobooster ! I'll try to check this out this weekend

flow-typed/npm/redux-loop_v4.x.x.js
```diff
@@ -0,0 +1,176 @@
...
+import type { ActionCreator, StoreEnhancer, Store } from "redux";
+
+declare module "redux-loop" {
+  declare export type StoreCreator = <S, A>(
+    reducer: LoopReducer<S, A>,
+    preloadedState: S,
+    enhancer: StoreEnhancer<S>
+  ) => Store<S>;
+
+  declare export type StoreCreatorExplicit<S, A> = (
```
bdwain: what is the difference between storecreator and storecreatorexplicit?

flow-typed/npm/redux-loop_v4.x.x.js
```diff
@@ -0,0 +1,176 @@
...
+    type: "RUN";
+    func: (...args: Array<C>) => B;
+    args?: Array<C>;
+    failActionCreator?: ActionCreator<A, any>;
+    successActionCreator?: ActionCreator<A, B>;
+    forceSync?: boolean;
+    simulate(simulation: CmdSimulation<B>): A;
+  }
+
+  declare export type SequenceCmd<A> = ListCmd<A>;
```
bdwain: i'd rather not export these since they're deprecated anyway. Since this is new, no need to support deprecated stuff

flow-typed/npm/redux_v4.x.x.js
bdwain: i'm confused why flow-typed would not work? What does it do then? Seems like you could add a post-install script to your package to install the flow types for the packages you want

flow-typed/npm/redux-loop_v4.x.x.js
stereobooster: ok. I mean this is direct translation of TS signatures

flow-typed/npm/redux-loop_v4.x.x.js
```diff
@@ -0,0 +1,176 @@
...
+    tagger: (subAction: B) => A
+  ): MapCmd<A>;
+  declare function Cmd$map<A, B, C>(
+    cmd: CmdType<B>,
+    tagger: (arg1: C, subAction: B) => A,
+    args?: Array<C>
+  ): MapCmd<A>;
+  declare function Cmd$map<A, B, C>(
+    cmd: CmdType<B>,
+    tagger: (arg1: C, arg2: C, subAction: B) => A,
```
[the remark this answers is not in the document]
stereobooster: if it would be in the end than yes - I could use rest

```ts
tagger: (subAction: B, ...rest: C[]) => A,
```

How much maximum arguments we want? Up to 9?

flow-typed/npm/redux-loop_v4.x.x.js
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  declare export interface NoneCmd {
+    type: "NONE";
+    simulate(): null;
+  }
+
+  declare export interface ListCmd<A> {
+    type: "LIST";
+    cmds: CmdType<A>[];
```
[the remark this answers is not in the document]
stereobooster: Again, direct translation of TS, I can remove everything except `type` and `simulate`

flow-typed/npm/redux-loop_v4.x.x.js
stereobooster: - `StoreCreator = <S, A>` will try to guess type from the context
- `StoreCreatorExplicit<S, A>` requires types explicitly

First one doesn't always work, so the second one is "plan b" option

flow-typed/npm/redux-loop_v4.x.x.js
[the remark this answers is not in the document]
bdwain: i don't see a reason to impose a limit. a rest parameter would allow any number

flow-typed/npm/redux-loop_v4.x.x.js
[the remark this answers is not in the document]
bdwain: yea i'm just realizing in hindsight that we maybe should not have exposed those things on typescript

flow-typed/npm/redux-loop_v4.x.x.js
bdwain: yea typescript was done before those were deprecated.

flow-typed/npm/redux-loop_v4.x.x.js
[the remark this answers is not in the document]
bdwain: Oh I see. because the rest parameter would need to be first? In that case, let's leave it at 3 and worry about adding more later.

flow-typed/redux-loop_v4.x.x.js
```diff
@@ -0,0 +1,176 @@
...
+    dispatch: Dispatch,
+    getState: GetState,
+    none: NoneCmd
+  };
+
+  declare export type ReducerMapObject<S, A> = {
+    [K: $Keys<S>]: LoopReducer<$Values<S>, A>
+  };
+
+  declare export function combineReducers<S, A>(
```
stereobooster:
```ts
  declare export type ExtractState = <S>(
    Reducer<S, any> | LoopReducer<S, any> | LiftedLoopReducer<S, any>
  ) => S;
  declare export type ExtractLiftedState = <S>(LiftedLoopReducer<S, any>) => S;
  declare export function combineReducers<O: Object, A>(
    reducers: O
  ): LiftedLoopReducer<$ObjMap<O, ExtractState>, A>;
```

flow-typed/npm/redux_v4.x.x.js
stereobooster: done

stereobooster:
I removed redux signatures. Didn't change anything about some stuff being public, I can mirror that as soon as TypeScript will be updated. Anything else I need to fix?

bdwain:
nope looks good! thanks!

PR #190, opened 2018-08-02
Add documentation to TS definitions

stereobooster:
This is not final version. Just an idea. WDYT?

screen shot 2018-08-02 at 16 38 51

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -85,13 +85,44 @@ declare function loop<S, A extends Action>(
 ): Loop<S, A>;
 
 declare namespace Cmd {
+  /**
+   * A symbol that can be passed to a cmd object as an arg (from a reducer)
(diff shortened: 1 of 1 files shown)


bdwain:
This is great! Thanks! I can merge this unless there was more you wanted to add.

stereobooster:
I can add the rest of it. I didn't do it because wanted to show it to you first. Another thing came to my mind, how we will make sure it will not get outdated. Maybe we need one source of truth from which we will generate all documentation jsdoc or something

bdwain:
I think for now we can just update the docs, and then worry about keeping it up to date separately. The library is not changing too quickly so I think it's fine to just try to make sure it gets updated with each pr.

Thanks!

bdwain:

PR #191, opened 2018-08-02
Some experiments to improve signatures

stereobooster:
Work in progress. Again not final version. Some experiments to improve signatures after experience with Flow

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -18,8 +18,8 @@ export interface LiftedLoopReducer<S, A extends Action> {
   (state: S | undefined, action: AnyAction, ...args: any[]): Loop<S, A>;
 }
 
-export type CmdSimulation = {
-  result: any,
+export type CmdSimulation<R = any> = {
+  result: R,
   success: boolean
 };
 export interface MultiCmdSimulation {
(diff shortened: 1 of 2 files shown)


bdwain:
Can you explain a bit more about what you're trying to improve? Also, are these breaking changes?

stereobooster:
> Can you explain a bit more about what you're trying to improve? 

I added comments in test files, like:

```diff
+ // checks third argument
```

This is what I changed. But I can write this list explicitly - small change log.

>  Also, are these breaking changes?

It is not breaking, but if people have some type errors which were unnoticed before they will become visible, so this can be perceived as breaking changes by some users

stereobooster:
Changelog:
- use [opaque type](https://flow.org/en/docs/types/opaque-types/) for `dispatch` and `getState`, to make sure people use exactly those values, not other symbols
- add intersection with functions to `dispatch` and `getState`, so they would work in cases like this `Cmd.run(dispatchNoop, { args: [Cmd.dispatch] })`, so `dispatchNoop` would know what params would be passed to it
- Add return type to `CmdSimulation`, so `simulate` function is aware about type of `result` type
- Add more types to `Cmd.run`, so know it makes sure that return value of the side-effect has the same type as input of `successActionCreator`
- Add more types to `Cmd.run`, so it checks that `args` fits as input for the side-effect function

Good enough? I'm ready to add more information is needed

index.d.ts
```diff
@@ -103,19 +107,35 @@ declare namespace Cmd {
...
+    args: C[]
+  ): MapCmd<A>;
+  export function map<A extends Action, B extends Action, C = any>(
+    cmd: CmdType<B>,
+    tagger: (arg1: C, arg2: C, subAction: B) => A,
+    args: C[]
+  ): MapCmd<A>;
+  export function map<A extends Action, B extends Action, C = any>(
+    cmd: CmdType<B>,
+    tagger: (arg1: C, arg2: C, arg3: C, subAction: B) => A,
```
bdwain: is it possible to do a ...rest param if it's not the last argument?

bdwain:
these all seem like good ideas. thanks!

index.d.ts
bdwain: Ok cool. I'm assuming no, and in that case then this is fine.

sbdchd:
Any update on this?

Adding types to ensure the `args` passed to the function of `Cmd.Run()` would be helpful.

```ts
function foo(a: string) {}

Cmd.run(
  foo, {
    args: ["bar"]
  }
)
```

bdwain:
@stereobooster are you still interested in making these changes? Do they conflict with #192 ?

bdwain:

PR #192, opened 2018-08-05
Better typings for Cmd.run (Typescript 3.0)

antoinewdg:
If you ever want to make the jump to Typescipt 3.0 (which I understand you may not want to), this PR enables better typings for Cmd.run:

- `option.args` has the same type as `f`'s arguments, and is mandatory if `f` takes any argument
- `otpions.successActionCreator` takes as only argument the return value of `f` (unwrapped if `f` returns a `Promise`)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -84,6 +84,10 @@ declare function loop<S, A extends Action>(
   cmd: CmdType<A>
 ): Loop<S, A>;
 
+type UnwrapPromise<T> = T extends Promise<infer U> ? U : T;
+type UnwrappedReturn<F> = F extends (...args: any[]) => infer R ? UnwrapPromise<R> : never; 
+type FunctionArgs<F> = F extends (...args: infer T) => any ? T : never;
+
 declare namespace Cmd {
   export const dispatch: symbol;
   export const getState: symbol;
@@ -107,12 +111,21 @@ declare namespace Cmd {
     args?: any[]
   ): MapCmd<A>;
 
-  export function run<A extends Action>(
-    f: Function,
+  export function run<A extends Action, F extends () => any>(
+    f: F,
     options?: {
-      args?: any[];
       failActionCreator?: ActionCreator<A>;
-      successActionCreator?: ActionCreator<A>;
+      successActionCreator?: (arg: UnwrappedReturn<F>) => A;
+      forceSync?: boolean;
+    }
+  ) : RunCmd<A>;
+
+  export function run<A extends Action, F extends Function>(
+    f: F,
+    options: {
+      args: FunctionArgs<F>;
+      failActionCreator?: ActionCreator<A>;
+      successActionCreator?: (arg: UnwrappedReturn<F>) => A;
       forceSync?: boolean;
     }
   ): RunCmd<A>;
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "rollup-plugin-node-resolve": "^3.0.0",
     "rollup-plugin-replace": "^2.0.0",
     "rollup-plugin-uglify": "^3.0.0",
-    "typescript": "^2.5.3",
+    "typescript": "^3.0.1",
     "typescript-definition-tester": "^0.0.6"
   },
   "dependencies": {},
--- a/yarn.lock
+++ b/yarn.lock
@@ -121,13 +121,13 @@ ansi-escapes@^3.0.0:
   version "3.1.0"
   resolved "https://registry.yarnpkg.com/ansi-escapes/-/ansi-escapes-3.1.0.tgz#f73207bb81207d75fd6c83f125af26eea378ca30"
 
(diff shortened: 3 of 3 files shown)


index.d.ts
```diff
@@ -107,12 +111,21 @@ declare namespace Cmd {
...
+      successActionCreator?: (arg: UnwrappedReturn<F>) => A;
+      forceSync?: boolean;
+    }
+  ) : RunCmd<A>;
+
+  export function run<A extends Action, F extends Function>(
+    f: F,
+    options: {
+      args: FunctionArgs<F>;
+      failActionCreator?: ActionCreator<A>;
```
stereobooster: `failActionCreator` is not typeable, for the same reason as Promise fail is not typeable.

Pseudo code:

```ts
class Promise<A> {
  constructor((resolve: (A) => void, reject: (any) => void) => void): void
}
```

This is because:

```ts
new Promise<number>((resolve, reject) => {
  throw "I can throw anything I want";
  // this is the same as
  reject("I can throw anything I want")
})
```

index.d.ts
```diff
@@ -84,6 +84,10 @@ declare function loop<S, A extends Action>(
   cmd: CmdType<A>
 ): Loop<S, A>;
 
+type UnwrapPromise<T> = T extends Promise<infer U> ? U : T;
+type UnwrappedReturn<F> = F extends (...args: any[]) => infer R ? UnwrapPromise<R> : never;
```
stereobooster: function in `Cmd.run` can return promise or result directly or throw exception.

stereobooster:
Take a look I did similar improvement https://github.com/redux-loop/redux-loop/pull/191

index.d.ts
antoinewdg: Yes, that's why I did not change its type. Is there something I did not see ?

antoinewdg:
@stereobooster I saw your PR, mine is a little different in scope because it requires TS 3.0, while yours can be used with the current version. Sorry if this looks like I'm "competing" with you, this is absolutely not what I intended :)

The goal of this PR is more to show that we could have some really better typings with this new version. Again, I know there are good arguments for not changing version (even if TS 3.0 has almost no breaking changes), and I just want to open the discussion.

antoinewdg:
By the way I think the `test/typescript.spec.ts` may not work as intended, because my changes definitely broke them and the tests are still green :(

stereobooster:
> By the way I think the test/typescript.spec.js may not work as intended, because my changes definitely broke them and the tests are still green :(

When I changed code, it failed here, because `dispatchNoop` expects function, not a symbol as an argument.

```ts
Cmd.run(dispatchNoop, { args: [Cmd.dispatch] })
```

bdwain:
Since this requires ts 3.0, it would be a breaking change right?

antoinewdg:
@bdwain yes it would be a breaking change because it would require users to use typescript 3.0.
I could rewrite it to support 2.8, which folks at DefinitelyTypes seem to think is reasonable: typings for most `react-*` require 2.8, as do some other popular libraries like `bluebird`.

It would allow the same guarantees as this PR, but be a little more verbose (which would be of no concern for users). Tell me what you guys think.

bdwain:
@antoinewdg yea if it's possible to support both that would be great I think. I'd rather avoid introducing a breaking change if possible

aeirola:
I'm also interested in seeing better typescript types for redux-loop, and have opted for overriding the types locally until the package types are improved. Luckily TypeScript 3.1 introduces [`typesVersions`](https://blogs.msdn.microsoft.com/typescript/announcing-typescript-3-1/#version-redirects-for-typescript-via-typesversions) with which you can provide different types for different versions of typescript. This way you would be able to introduce better types, without breaking backwards compatibility with older versions of typescript. Managing multiple different typings files might become tedious though.

bdwain:
@antoinewdg sounds like v3.1 might help this.

bdwain:

sbdchd:
I think this would be nice and I'd be happy to open another PR with the same functionality if necessary.

These types would have prevented the following error in the typescript example linked in the docs:

https://github.com/meandmax/redux-loop-ts-example/pull/1

which is caused by the `successActionCreator` args not agreeing with the function passed to `Cmd.Run`

bdwain:
@sbdchd thanks! I'd be happy to look at a PR if you open it.

bdwain:
since this one is out of date maybe easiest to open a new one?

bdwain:
I see now you opened #207

sbdchd:
#207 actually covers a different typing case, allowing for success and failure actions to be different types in the union of actions.

This pr ensures that `args` and the params of the function passed to `Cmd.run()` match.

bdwain:
got it. well if you want to reopen this one in a new PR and with the comments addressed I'll be happy to check it out.

PR #194, opened 2018-08-08
fix error swallowing for cmd.run

jamesjtong:
Have redux loop log out errors that occurred within sideEffect functions instead of swallowing them in Cmd.run.

To reproduce the issue, throw an error inside your sideEffect function and notice that the error is being swallowed (not being showed in console). This makes debugging more difficult and hides errors from monitoring services.

### DIFF
--- a/src/cmd.js
+++ b/src/cmd.js
@@ -32,14 +32,15 @@ function handleRunCmd(cmd, dispatch, getState){
     let result = cmd.func(...getMappedCmdArgs(cmd.args, dispatch, getState))
 
     if (isPromiseLike(result) && !cmd.forceSync){
-      return result.then(onSuccess, onFail).then(action => {
-        return action ? [action] : [];
+      return result.then( onSuccess, (error) => {
+        console.error(error);
+        return onFail(error);
       })
+      .then(action => action ? [action] : [])
     }
     let resultAction = onSuccess(result);
     return resultAction ? Promise.resolve([resultAction]) : null;
-  }
-  catch(err){
+  } catch(err){
     if(!cmd.failActionCreator){
       console.error(err);
       throw err //don't swallow errors if they are not handling them
--- a/test/cmd.spec.js
+++ b/test/cmd.spec.js
@@ -45,11 +45,14 @@ describe('Cmds', () => {
         });
 
         it('resolves with an empty array if the function returns a rejected promise', async () => {
+          let consoleErr = jest.spyOn(console, 'error').mockImplementation(() => {});
           sideEffect.mockReturnValueOnce(Promise.reject(123));
           let cmd = Cmd.run(sideEffect);
           let result = executeCmd(cmd, dispatch, getState);
           expect(sideEffect.mock.calls.length).toBe(1);
           await expect(result).resolves.toEqual([]);
+          expect(consoleErr).toHaveBeenCalled();
+          consoleErr.mockRestore();
         });
 
         it('rethrows the thrown error if the function throws', function(){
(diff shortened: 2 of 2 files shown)


src/cmd.js
```diff
@@ -32,7 +32,14 @@ function handleRunCmd(cmd, dispatch, getState){
     let result = cmd.func(...getMappedCmdArgs(cmd.args, dispatch, getState))
 
     if (isPromiseLike(result) && !cmd.forceSync){
-      return result.then(onSuccess, onFail).then(action => {
+      return result.then(
+        onSuccess,
+        (error) => {
```
stereobooster: if it would be development build and production build like in React 🤔

jamesjtong: Sorry, I don't quite understand. 

I think the console should still be showing errors in both development and production.

This PR could help with issues like https://github.com/redux-loop/redux-loop/issues/195 where errors are disappearing into the void.

stereobooster: it can create a lot of noise. I would say if developer provided handler than we do not need to log it.

```js
onFail = cmd.failActionCreator || (e => console.error(e))
```

bdwain: Hm I see what @stereobooster is saying. Maybe this could be a configuration option? Whether to log handled errors? And then always log unhandled errors?

jamesjtong: My thoughts are that the warnings that React provides in development mode help alert issues related to **possible** app breakage and performance (ex: not specifying keys in lists or incorrect Proptypes). The errors that redux-loop currently hides are native javascript errors that I would think most developers would want to know about (either through the console or through monitoring services)

I think a configuration option is fair to support the users of the library who don't want the library to log the errors out by default.

But then comes again the question of what the default should be? Should we by default, 
a) leave it off by default, continue swallowing errors, and keep existing behavior for consumers who upgrade
b) have it on by default and log errors for code that breaks. It is then up to the consumer to catch these errors within the failHandler if they truly want to hide them.

I am assuming the majority of users would want b?

What are your thoughts? @bdwain , @stereobooster

jamesjtong: That is true too. In the codebase where I use redux-loop, there are failActionCreators on all requests to the API. I originally ran into this issue because there were no console.error(error) within those failActionCreators.

I think the disadvantage of this approach,
```(onFail = cmd.failActionCreator || (e => {console.error(e)}))``` 
, is that the consumers who already defines failActionCreators will have to add those console.error lines to their failActionCreators or an abstraction to handle it. 
In addition, for new users to the library, will they know to provide it without diving into their code to know why their errors are being swallowed?

This is also a question if we want to inconvenience the consumer who already has a failActionCreator and who wants to see the errors but doesnt have the console.error to add that additional line of console.error to their failActionCreators OR inconvenience the other percentage of consumers who don't want to see errors in the console to catch it.

I am starting to think that the configuration option with the default is ideal as it can handle both options without adding extra code for both types of consumers.

bdwain: I think the question we need to ask is whether things running in failActionCreator are meant to be exceptional circumstances (like the fact that they run at all means something is wrong ) or whether they are normal app flow. If it’s the former, I think we want To always log (rather than expect you to manually log in every error handler). If not, we probably don’t want to log (or log only if logging is enabled).

From how I’ve used redux loop, they typically only run when something went wrong, so I’m starting to lean towards always logging. I kind of like doing what react does, which is log it only if in dev mode. No need to explicitly configure. Though i suppose we could add an option to log even in prod mode, though I im not sure that’s necessary. 

Thoughts?

bdwain: Actually the more that I think about it, I think react only does dev mode because there’s a perf hit when checking proptypes. And those are more warnings than errors. 

We already know we’re in an error state and there won’t Be a perf hit. So maybe we just want to always log? In which case maybe a config option to log only unhandled errors would be nice if someone really doesn’t want it. But by default I’d say log everything. 

I guess that was what you said @jamestong. So yea I like that approach.

jamesjtong:
added a config option with a key of LOG_ERRORS_ON_HANDLED_FAILURES and a default of true

The entry point that I chose for config was install. Please advise if there is a better entry point or any other suggestions.

Thanks

bdwain:
@jamesjtong install seems fine to me. And I agree it seems like a lot of config. Maybe we only add that if people are unhappy about the logging? I'm fine with that. In general I think failure handlers are exceptional cases so it's ok to log them. Better than swallowing things you don't want to miss.

jamesjtong:
Yea that makes sense to me too. I have squashed the commit and pushed up changes to just always log when the effect has an error (pretty much back to my first commit). I did look into removing that try/catch block some more and tried removing it, but then realized that it was necessary for https://github.com/redux-loop/redux-loop/blob/8c9631cf7ab92f961d1d2d86db55ec5529ede1bb/test/cmd.spec.js#L133 which is catching against the cmd.func execution to get the "result" variable.

If people are unhappy about the logging and the config param to install is ever needed, then that is here (https://github.com/redux-loop/redux-loop/compare/master...jamesjtong:jt/fix-issue-with-error-swallowing-with-configs-save).

bdwain:
looks good! Thanks!

PR #196, opened 2018-09-06
Removed console.error on command promise failure if failActionCreator exists

dmotz:
`4.4.1` introduced a bug where `console.error` is always called on command promise failures, even if a `failActionCreator` exists.

This isn't particularly impeding on web projects, but is a showstopper in React Native projects since `console.error` throws up a red fullscreen error message UI.

### DIFF
--- a/src/cmd.js
+++ b/src/cmd.js
@@ -33,7 +33,9 @@ function handleRunCmd(cmd, dispatch, getState){
 
     if (isPromiseLike(result) && !cmd.forceSync){
       return result.then( onSuccess, (error) => {
-        console.error(error);
+        if(!cmd.failActionCreator){
+          console.error(error);
+        }
         return onFail(error);
       })
       .then(action => action ? [action] : [])


bdwain:
This was intentional (see discussion in #194), though I was not aware of the behavior of console.error on react native. Sorry about that. Is there an alternative log function available in react native that is not so blatant? if so, we could maybe make the logging function a config parameter and use that, rather than console.error (and default to console.error)

If not, I think we still want to have the logging enabled by default (see the discussion in the long thread in that PR), but we can bring back the option to disable logging of handled errors.

dmotz:
It seems the current behavior is inconsistent. If an error is thrown synchronously, `console.error` is only called if a `failActionCreator` isn't handling the error. However, for async errors (promises), the presence of a `failActionCreator` is ignored and `console.error` is called regardless. This PR addresses that inconsistency.

Reference: https://github.com/redux-loop/redux-loop/blob/master/src/cmd.js#L44

bdwain:
Yes, but you're changing the behavior of the asynchronous case, when the synchronous case is the one that is being handled incorrectly. 

Would you be interested in doing a PR to add an option to disable logging of handled errors? I think that would address your issue. And we should also move the console.error on line 44 to outside of the `if` that it's in.

dmotz:
I guess I misunderstood the aim of #194. Personally I wouldn't expect the library to produce its own side effects (`console.error`) when I've designated my own explicit error handlers.

Regarding the config option, it looks like that's already been setup here: https://github.com/redux-loop/redux-loop/compare/master...jamesjtong:jt/fix-issue-with-error-swallowing-with-configs-save

bdwain:
Yea the goal was to make errors more noticeable. [This comment](https://github.com/redux-loop/redux-loop/pull/194#discussion_r210234647) kind of sums up how we landed on it.

> I think the question we need to ask is whether things running in failActionCreator are meant to be exceptional circumstances (like the fact that they run at all means something is wrong ) or whether they are normal app flow. If it’s the former, I think we want To always log (rather than expect you to manually log in every error handler). If not, we probably don’t want to log (or log only if logging is enabled).

> From how I’ve used redux loop, they typically only run when something went wrong, so I’m starting to lean towards always logging. I kind of like doing what react does, which is log it only if in dev mode. No need to explicitly configure. Though i suppose we could add an option to log even in prod mode, though I im not sure that’s necessary.

 And yea we originally made an option to disable it. Feel free to copy most of that work.

stereobooster:
If developer provided failAction creator in means they know that error can happen, they already handle it. I don't see why you need visibility for the error that you already handling. If you use `.catch` in promise there is no warning if promise get rejected, (but there is one if you don't). If you use try-catch around code which throws exception there is no warning in the console, but there is one if nothing catched exception.

jamesjtong:
hmm yes perhaps commit a0e772177b1ea8 should be reverted. Only pain point would be adding those console.errors in the failActionHandlers for the developers who already use failActionHandlers but don't explicity log it.

The config option is quite verbose. Not sure if it is worth it. ¯\_(ツ)_/¯

bdwain:
I'd rather err on the side of warning more than less. But I'm happy to accept a PR adding the option to disable logging for a handled error. That would address everyone's concerns and still give a default behavior that makes sure you are aware when things are going wrong.

dmotz:
Closing in favor of #198.

PR #197, opened 2018-09-06
documentation was wrong on compose liftstate usage

thewillhuang:
Fixes the documentation for redux-loop lift state and compose usage


bdwain:
Good catch! Thanks!

PR #198, opened 2018-10-06
add config to control logging of effect errors even when handled

jamesjtong:
Add a config option to install.

 Currently the only option is `DONT_LOG_ERRORS_ON_HANDLED_FAILURES` (see https://github.com/redux-loop/redux-loop/pull/194, which fixed the issue where error messages were being swallowed in effects when the client added a failActionHandler on a Cmd) which is set to false by default. This means that by default, redux-loop will log errors in your effects even if you pass in a failActionCreator to a command (passing in a failActionCreator does not swallow errors like how `catch` is used for Promises).

### DIFF
--- a/docs/api-docs/install.md
+++ b/docs/api-docs/install.md
@@ -1,4 +1,11 @@
-# `install`
+# `install(config)`
+
+* `config: Object<string, Boolean>` &ndash; an optional map of config values to override default
+ redux-loop configs
+*  Currently the only option is `DONT_LOG_ERRORS_ON_HANDLED_FAILURES` which is set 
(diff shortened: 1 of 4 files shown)


bdwain:
Sorry for the delay. This looks good. Thanks!

2 things:
Can the option be flipped and be false by default? DONT_LOG_ERRORS_ON_HANDLED_FAILURES? I think options that are false by default are more natural.

Also, can you add some documentation for this option? Probably just something quick in the install page. https://redux-loop.js.org/docs/api-docs/install.html (the docs are in this repo).

Thanks!

bdwain:
Sorry for the delay! Thanks for fixing this!

PR #199, opened 2018-10-19
Add some test cases for reduce and combine reducers working together

g12i:
Hello there 👋 

I was in doubt whether the `reduceReducers` and `combineReducers` works together so I figured out the best way to check it would be to write some tests. 

Feel free to close this PR if you find them pointless.


bdwain:
more tests never hurts! Thanks!

PR #200, opened 2018-10-24
Batch illustration

g12i:
Hello! 

It took me a while to understand the difference between `batch` option in `Cmd.list`. I know that it is going to be deprecated in 5.0.0. But until then maybe this illustration will help the others (who has the same problem) understand it.

⭐️ Preview: https://github.com/redux-loop/redux-loop/blob/55646d808b698cb2e8cfe3f0e41932c76f7b3076/docs/api-docs/cmds.md#batch-option


bdwain:
Thanks for the contribution! Sorry it took me so long to merge it.

PR #202, opened 2018-11-28
Fix: Show addition result of addReducer in docs

sdaves:
The result of addReducer should be to add the change value (6 in this case) to the state's add property (5 in this case) so it would be 11 and not 30 like you would get from multiplying.


bdwain:
good catch. thanks!

Issue #203, opened 2018-12-04
Add install function's config options to type definitions


mllustosa:
I just updated to version `4.5.0` and noticed that I cannot use the new config options (when installing the middleware) to add `DONT_LOG_ERRORS_ON_HANDLED_FAILURES`, because it's lacking the proper TS definition.

It would be great if you guys could update the type definitions to allow the use of the option.

Thanks a lot in advance and keep up the great work!

bdwain:
closed via #204

PR #204, opened 2018-12-04
Update type definition of install function to include optional config object

mllustosa:
This adds an optional config object to the install function. It's important because it allows people who are using on Typescript to use the config options.

Based on issue #203


bdwain:
sorry i missed this!

PR #205, opened 2018-12-19
Have dispatch return the action it's provided after the current loop concludes

sasucker:
The vanilla implementation of redux [passes the provided action back](https://github.com/reduxjs/redux/blob/628ffdd4c5a1b2e4e5b42b26ec8f7b29c344bf0c/src/createStore.js#L208) as a convenience. [The doc comment](https://github.com/reduxjs/redux/blob/628ffdd4c5a1b2e4e5b42b26ec8f7b29c344bf0c/src/createStore.js#L173-L174) does mention that this isn't guaranteed behavior across middleware/store enhancers, but during our transition to `redux-loop` from `redux-thunk`, this characteristic of `dispatch` is important because it was pretty heavily relied on in the past.

Just looking at the code, it's unclear why we do *not* return the provided action. Especially if there are multiple actions in the [cmdRun queue](https://github.com/redux-loop/redux-loop/blob/d811096459104856b4bb7eaf8e8f641b1c23c3fd/src/install.js#L32), we seem to discard the result of the Promises. 

Applying this change has not broken our limited usage of `redux-loop`, and I'm curious if it may break in other places.

Thanks!

### DIFF
--- a/src/install.js
+++ b/src/install.js
@@ -50,10 +50,10 @@ export function install(config={}) {
     }
 
     const dispatch = (action) => {
-      store.dispatch(action)
+      const result = store.dispatch(action)
       const cmdsToRun = cmdsQueue
       cmdsQueue = []
-      return runCmds(cmdsToRun)
+      return runCmds(cmdsToRun).then(() => result);
     }
 
     const replaceReducer = (reducer) => {
--- a/test/index.spec.js
+++ b/test/index.spec.js
@@ -114,4 +114,41 @@ describe('redux loop store enhancer', () => {
       done();
     });
   });
+
+  describe('dispatch', () => {
+    it('should return the provided action after the dispatch promise resolves', (done) => {
+      const firstAction = {type: 'FIRST_ACTION'};
+      const downstreamAction = {type: 'DOWNSTREAM_ACTION'};
+      const thunk = (dispatch) => Promise.resolve(dispatch(downstreamAction));
+      const initialState = {
+        firstRun: false,
+        secondRun: false,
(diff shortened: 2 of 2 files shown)


bdwain:
Hi @sasucker. Thanks for the PR.

Can you explain a little more what you need the return value for? Especially because vanilla redux returns the value synchronously and this would return asynchronously, it seems like you'd have to change all of your code anyway.

sasucker:
Hi @bdwain, thanks for getting back so quickly!

As mentioned, we previously (and currently) use `redux-thunk` for our async behavior. For better or for worse, some of our older components rely on the underlying `Promise` that is returned from an async thunk.

To illustrate:
```jsx
function asyncAction() {
    return (dispatch) => {
        dispatch(fetching());
        return request().then((response) => 
            dispatch(complete(response.data));
        );
    };
}

// although in actuality we'd use `bindActionCreators`, which also returns the `dispatch`
class SomeComponent {
    componentDidMount() {
        this.setState({loading: true});

        this.props.store.dispatch(asyncAction()).then((datas) => {
            this.setState({datas, loading: false});
        });
    }

    render() {
        return this.state.loading
            ? <h1>{this.state.datas}</h1>
            : <LoadingIndicator />;
    }
};
```
We've largely moved away from this pattern, but in the interest of transitioning it would definitely help us if `redux-loop`'s dispatch also returned the provided action.

bdwain:
Got it. Thanks for explaining. I don't see this being a problem since we currently always return undefined.

Sorry I couldn't get back sooner. I was pretty busy over the holidays.

PR #206, opened 2019-01-11
Relax reduceReducers typing to match implementation

aeirola:
Previously the typescript types for the `reduceReducers` function required all of the reducers to handle the initial `undefined` state, although the actual implementation ensures that only the first reducer to only ever receive an `undefined` state. The types thus forced all the reducers to implement some unnecessary handling for the initial state.

This PR allows all but the first reducer to assume that the state they receive is always initialised.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -14,7 +14,7 @@ export interface LoopReducer<S, A extends Action> {
   (state: S | undefined, action: AnyAction, ...args: any[]): S | Loop<S, A>;
 }
 
-export interface NonInitialLoopReducer<S, A extends Action> {
+export interface LoopReducerWithDefinedState<S, A extends Action> {
   (state: S, action: AnyAction, ...args: any[]): S | Loop<S, A>;
 }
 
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -14,6 +14,10 @@ export interface LoopReducer<S, A extends Action> {
   (state: S | undefined, action: AnyAction, ...args: any[]): S | Loop<S, A>;
 }
 
+export interface NonInitialLoopReducer<S, A extends Action> {
```
aeirola: I'm not quite happy with the naming of this interface. Suggestions for better names are welcome.

bdwain: LoopReducerWithDefinedState?

bdwain:
Good call! Thanks!

I had a naming suggestion for NonInitialLoopReducer. Other than that, seems good.

index.d.ts
aeirola: Yeah, that does feel more consistent. I'll go ahead and change it.

aeirola:
Fixed the type name. I'm happy with the result as it is.

PR #207, opened 2019-02-01
Update TypeScript types for Cmd.run()

sbdchd:
## Before

When using defining action creators that return their specific action instead of the union of all actions, there would be a type error with `Cmd.run`

## After

Action creators that return different actions in the union of actions can be using by `Cmd.run()`

## Example

Here is an example that breaks with the current typings:

```typescript
interface IFetchFooRequest {
  type: typeof FETCH_FOO_REQUEST
}

interface IFetchFooSuccess {
  type: typeof FETCH_FOO_SUCCESS
}

const fetchFooSuccess = (): IFetchFooSuccess => ({
  type: FETCH_FOO_SUCCESS,
})

interface IFetchFooFailure {
  type: typeof FETCH_FOO_FAILURE
}

const fetchFooFailure = (): IFetchFooFailure => ({
  type: FETCH_FOO_FAILURE,
})

// --snip--
    case FETCH_FOO_REQUEST:
      return loop(
        state,
        Cmd.run(apiFetchFoo, {
          successActionCreator: fetchFooSuccess,
          failActionCreator: fetchFooFailure,
    // [ts]
    // Argument of type '{ successActionCreator: () => IFetchFooSuccess; failActionCreator: () => IFetchFooFailure; }' is not assignable to parameter of type '{ args?: any[] | undefined; failActionCreator?: ActionCreator<IFetchFooFailure> | undefined; successActionCreator?: ActionCreator<IFetchFooFailure> | undefined; forceSync?: boolean | undefined; }'.
    //   Types of property 'successActionCreator' are incompatible.
    //     Type '() => IFetchFooSuccess' is not assignable to type 'ActionCreator<IFetchFooFailure>'.
    //       Type 'IFetchFooSuccess' is not assignable to type 'IFetchFooFailure'.
    //         Types of property 'type' are incompatible.
    //           Type '"FETCH_FOO_SUCCESS"' is not assignable to type '"FETCH_FOO_FAILURE"'. [2345]

// --snip--
```

This isn't an issue when typing action creators to return the union `Action`. However, if you use something like [`typesafe-actions`](https://github.com/piotrwitek/typesafe-actions) which constructs the union of actions using the `ReturnType` of action creators there are type errors.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -115,12 +115,12 @@ declare namespace Cmd {
     args?: any[]
   ): MapCmd<A>;
 
-  export function run<A extends Action>(
+  export function run<A extends Action, B extends Action>(
     f: Function,
     options?: {
       args?: any[];
       failActionCreator?: ActionCreator<A>;
-      successActionCreator?: ActionCreator<A>;
+      successActionCreator?: ActionCreator<B>;
       forceSync?: boolean;
     }
   ): RunCmd<A>;
--- a/test/typescript/loopReducer.ts
+++ b/test/typescript/loopReducer.ts
@@ -9,6 +9,10 @@ import {
 } from '../../index';
 import { AnyAction } from 'redux';
 
+const FETCH_FOO_REQUEST = 'FETCH_FOO_REQUEST'
+const FETCH_FOO_SUCCESS = 'FETCH_FOO_SUCCESS'
+const FETCH_FOO_FAILURE = 'FETCH_FOO_FAILURE'
+
 type TodoState = { todos: string[]; nestedCounter: number };
 
 type TodoActions =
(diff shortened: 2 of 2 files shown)


index.d.ts
```diff
@@ -115,12 +115,12 @@ declare namespace Cmd {
     args?: any[]
   ): MapCmd<A>;
 
-  export function run<A extends Action>(
+  export function run<A extends Action, B extends Action>(
```
bdwain: so is this a breaking change? since it has 2 type params now.

sbdchd: Maybe. 

If someone is explicitly passing a type param then they will get a type error saying that they didn't pass enough type params.

But if they are letting typescript infer `A` then no change will be necessary.

```typescript
// this works with the change
case FETCH_FOO_REQUEST:
       return loop(
         state,
         Cmd.run(apiFetchFoo, {
           successActionCreator: fetchFooSuccess,
           failActionCreator: fetchFooFailure,
         })
       )
```
```typescript
// compiler errors
case FETCH_FOO_REQUEST:
       return loop(
         state,
         // typescript: Expected 2 type arguments, but got 1.
         Cmd.run<Action>(apiFetchFoo, {
           successActionCreator: fetchFooSuccess,
           failActionCreator: fetchFooFailure,
         })
       )
```

bdwain: hm alright. to be safe, I'll probably make this a major bump. I might try to include some other breaking changes before releasing it. It shouldn't be too long before i do though.

Thanks!

PR #208, opened 2019-02-12
Fix StoreCreator TypeScript def to match redux

kumar303:
This patch fixes the TypeScript definition of `StoreCreator` to make `preloadedState` optional.

Why? I'll explain.

Here is a TypeScript version of the documented way to create a store:

```js
import { createStore } from 'redux';
import { StoreCreator, install } from 'redux-loop';
import reducer from './reducer';

type ApplicationState = {};
const initialState: ApplicationState = { /* ... */ };

const enhancedCreateStore = createStore as StoreCreator;
const store = enhancedCreateStore(
  reducer, initialState, install<ApplicationState>()
);
```

However, `initialState` is optional in [createStore](https://redux.js.org/api/createstore). This should be possible:

```js
const store = enhancedCreateStore(
  reducer, undefined, install<ApplicationState>()
);
```

Yet it produces an error that looks something like this. The actual error is generated from the redux-loop test suite before my patch.

```
test/typescript/loopReducer.ts:174:3 - error TS2345: Argument of type 'LoopReducer<TodoState, TodoActions>' is not assignable to parameter of type 'LoopReducer<TodoState | undefined, TodoActions>'.
  Type 'TodoState | undefined' is not assignable to type 'TodoState'.
    Type 'undefined' is not assignable to type 'TodoState'.

174   todosReducer, undefined, enhancer
      ~~~~~~~~~~~~
```

It's very cryptic but I've tracked it down to the type definition of StoreCreator, which I've fixed to look more like [that of `redux`](https://github.com/reduxjs/redux/blob/d53364c44b2fb75b59e2c98090b253c103d63c75/index.d.ts#L232-L236).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3,7 +3,7 @@ import { Action, ActionCreator, AnyAction, StoreEnhancer, Store } from 'redux';
 export interface StoreCreator {
   <S, A extends Action>(
     reducer: LoopReducer<S, A>,
-    preloadedState: S,
+    preloadedState: S | undefined,
     enhancer: StoreEnhancer<S>
   ): Store<S>;
 }
@@ -84,7 +84,7 @@ export type CmdType<A extends Action> =
 export interface LoopConfig {
   readonly DONT_LOG_ERRORS_ON_HANDLED_FAILURES: boolean;
 }
-  
+
 declare function install<S>(config?: LoopConfig): StoreEnhancer<S>;
 
 declare function loop<S, A extends Action>(
--- a/test/typescript/loopReducer.ts
+++ b/test/typescript/loopReducer.ts
@@ -9,7 +9,7 @@ import {
   LoopReducer,
   StoreCreator,
 } from '../../index';
-import { AnyAction, compose, createStore } from 'redux';
+import { AnyAction, createStore } from 'redux';
 
 const FETCH_FOO_REQUEST = 'FETCH_FOO_REQUEST'
 const FETCH_FOO_SUCCESS = 'FETCH_FOO_SUCCESS'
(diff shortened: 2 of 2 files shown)


index.d.ts
```diff
@@ -3,7 +3,7 @@ import { Action, ActionCreator, AnyAction, StoreEnhancer, Store } from 'redux';
 export interface StoreCreator {
   <S, A extends Action>(
     reducer: LoopReducer<S, A>,
-    preloadedState: S,
+    preloadedState: S | undefined,
```
bdwain: good catch! Thanks! can this be

`preloadedState?: S`

instead?

kumar303: I did try that at first but it doesn't work because the question mark makes it an optional argument--rather than one defined explicitly with an undefined value--and you can't follow an optional argument by a required argument:

```
index.d.ts:7:5 - error TS1016: A required parameter cannot follow an optional parameter.

7     enhancer: StoreEnhancer<S>
      ~~~~~~~~
```

However, I could make them both optional which would follow the [`redux` definition](https://github.com/reduxjs/redux/blob/d53364c44b2fb75b59e2c98090b253c103d63c75/index.d.ts#L232-L236) more closely. Should I define it like this?

```diff
diff --git a/index.d.ts b/index.d.ts
index fa84b3e..9b560f2 100644
--- a/index.d.ts
+++ b/index.d.ts
@@ -3,8 +3,8 @@ import { Action, ActionCreator, AnyAction, StoreEnhancer, Store } from 'redux';
 export interface StoreCreator {
   <S, A extends Action>(
     reducer: LoopReducer<S, A>,
-    preloadedState: S | undefined,
-    enhancer: StoreEnhancer<S>
+    preloadedState?: S,
+    enhancer?: StoreEnhancer<S>
   ): Store<S>;
 }
 
```

The reason I didn't go that way is because the `install()` enhancer is required to make it all work.

bdwain: got it. that makes sense. thanks for explaining

bdwain:
this is good! thanks for making the fix! I'll merge it now, though there is an unreleased breaking change in master and I have a few things I'd like to do before releasing the new version, so I may not get to release this for a few days.

kumar303:
Thanks! No rush from my end.

PR #210, opened 2019-02-17
Better typings for reducers

kosmogradsky:
`Reducer` in Redux is defined like so:
```typescript
export type Reducer<S = any, A extends Action = AnyAction> = (
  state: S | undefined,
  action: A
) => S
```
I think it would be better to define `LoopReducer` in a similar way so the action type is taken from the parameter and the parameter has the same default as `Reducer`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -10,16 +10,16 @@ export interface StoreCreator {
 
 export type Loop<S, A extends Action> = [S, CmdType<A>];
 
-export interface LoopReducer<S, A extends Action> {
-  (state: S | undefined, action: AnyAction, ...args: any[]): S | Loop<S, A>;
+export interface LoopReducer<S, A extends Action = AnyAction> {
+  (state: S | undefined, action: A, ...args: any[]): S | Loop<S, A>;
 }
 
-export interface LoopReducerWithDefinedState<S, A extends Action> {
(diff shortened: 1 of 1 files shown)


bdwain:
but reducers are called with all actions? Why would you ever provide a different type there?

kosmogradsky:
You would want this to provide more exact type for better type inference in the reducer `switch` case. For example, in your app you can have:
```
type Action = {
  type: 'ACTION_ONE',
  payload: number
} | {
  type: 'ACTION_TWO',
  payload: string
} | {
  type: 'ACTION_THREE',
  payload: {
    property1: number,
    property2: string
  }
}
```
If you provide `Action` type to `Reducer` as a type parameter you get better type inference for payloads.

bdwain:
But it still needs to accept other action types for the default case,
right? How does that work

kosmogradsky:
You would still have a default case that returned state unchanged. If all the members of `Action` union type are handled with respectful `case`s then the default case will have the `action` typed as `never`.

bdwain:
sorry for the delay. I was on vacation the last few weeks. Sounds good!

bdwain:
sorry for the delay. I was on vacation the last few weeks. Sounds good!

PR #213, opened 2019-03-15
Make Cmd.run successActionCreator type-safe

felixeichhorst-jimdo:
Using redux-loop extensively, I want to rely on TypeScript checking if the argument for any `successActionCreator` in a RunCmd has actually the type of the return value of `f` or `f`s returned promise.
In order to do so, we need to refine the type definitions of `run`

Additionally, we need to set the node version to 11.10.1 for now to overcome an issue with TravisCI, jest and Node 11.11. See https://github.com/facebook/create-react-app/issues/6591.


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "node"
+  - "11.10.1" # Due to a bug with 11.11  and TravisCI
```
bdwain: sorry i took so long to respond to this. looks like this issue is fixed now so we probably don't need to hard code the node version anymore

index.d.ts
```diff
@@ -115,12 +115,12 @@ declare namespace Cmd {
...
 
-  export function run<A extends Action, B extends Action>(
-    f: Function,
+  export function run<A extends Action, B extends Action, R>(
+    f: (...args: any[]) => R | Promise<R>,
     options?: {
       args?: any[];
-      failActionCreator?: ActionCreator<A>;
-      successActionCreator?: ActionCreator<B>;
+      failActionCreator?: (...args: any[]) => A;
```
bdwain: can't the failActionCreator remain `ActionCreator<A>` ?

bdwain:
Sorry I wasn't able to look at this sooner. I was out of town for a bit and let a bunch of things pile up.

Overall, this change makes sense to me. I just had those 2 small comments, and also, can you rebase and squash and get rid of the merge commit? Thanks!

PR #216, opened 2019-05-09
feat(ts): improve run signature typings

felixeichhorst-jimdo:
Sorry for the inconvenience, we accidentally closed the [previous pull request](#213) by reseting the forks master to upstream.

We refined the changes from the previous PR incorporating the changes done in 268c98191bff6963795fda166376aa9b14106f19.

This makes sure that the result of the function passed to `run` is compatible with the arguments that the `successActionCreator` receives.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,13 @@
 import { Action, ActionCreator, AnyAction, StoreEnhancer, Store } from 'redux';
 
+export type PromiseResult<T> = T extends Promise<infer U> ? U : never;
+
+export type FinalReturnType<T extends (...args: any) => any> = T extends (
+  ...args: any
+) => Promise<any>
+  ? PromiseResult<ReturnType<T>>
+  : ReturnType<T>;
+
 export interface StoreCreator {
   <S, A extends Action>(
     reducer: LoopReducer<S, A>,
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -1,5 +1,13 @@
 import { Action, ActionCreator, AnyAction, StoreEnhancer, Store } from 'redux';
 
+export type PromiseResult<T> = T extends Promise<infer U> ? U : never;
```
bdwain: This looks a lot more complex than your original change. Can you explain why you needed to change so much? I'm not that familiar with typescript myself and I've mostly been relying on contributors to keep the types correct. I figured the only change you needed was the one I mentioned in the original PR.

index.d.ts
```diff
@@ -115,16 +123,22 @@ declare namespace Cmd {
...
-  export function run<A extends Action, B extends Action>(
-    f: Function,
+  export function run<
+    TSuccessAction extends Action,
+    TFailAction extends Action,
+    TFunction extends (...args: any[]) => any | Promise<any>
+  >(
+    f: TFunction,
     options?: {
       args?: any[];
```
aeirola: Thanks for the improvement to the types. I've been waiting for this for a long time. I'd suggest a small addition to the types:

The introduction of the `TFunction` type argument could also enable better type safety of the function arguments. So the type for the `args` option would be better defined as:

```typescript
  args?: Parameters<TFunction>;
```

bdwain:
closing this PR due to inactivity. @aeirola if you want to reopen it with the comment addressed I could try to take a look.

aeirola:
I don't think there is a need to reopen this, as #222 already seems to address the comments I left here. I think it is best to continue the work or improving the run signature typings there.

PR #217, opened 2019-05-20
unique symbol type for Cmd.getState, Cmd.dispatch

laszlopandy:
Typescript 2.7 added `unique symbol` to describe symbol types which are unique. Otherwise the types of Cmd.getState and Cmd.dispatch are assignable to each other.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -93,8 +93,8 @@ declare function loop<S, A extends Action>(
 ): Loop<S, A>;
 
 declare namespace Cmd {
-  export const dispatch: symbol;
-  export const getState: symbol;
+  export const dispatch: unique symbol;
+  export const getState: unique symbol;
   export const none: NoneCmd;
   export function action<A extends Action>(action: A): ActionCmd<A>;
   export function batch<A extends Action>(cmds: CmdType<A>[]): BatchCmd<A>;


bdwain:
Is this backwards compatible? Does this raise the minimum typescript version? Or are 2.7 features already being used?

laszlopandy:
These are good questions.
No `unique symbol` is not backwards compatible. It will cause an syntax exception for anyone using an earlier version of typescript.

I'm not sure if any other syntax being used requires >=2.7. The only way to ensure that would be to downgrade the typescript version in `package.json` and run the tests. Therefore, I assumed that the minimum version is what is specified in the package.json (typescript 3.2.2). Is a minimum version specified somewhere else?

laszlopandy:
If it is important to maintain backwards compatibility, it is possible to specify multiple `index.d.ts` files which will be used by different versions of the typescript compiler:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions

laszlopandy:
@bdwain I tested a range of typescript versions and it seems that `redux` is already using syntax that requires `>=2.8`. So it is safe to say that `unique symbol` does not raise the minimum typescript version for `redux-loop`.

bdwain:
@laszlopandy great sorry for the delay thanks for looking into that

PR #218, opened 2019-07-17
Cmd.map() wraps actions send via dispatch()

laszlopandy:
This was implemented inconsistently because actions returned from a command would be wrapped with the tagger but actions send through dispatch() would not.

I wrote an example program to demonstrate the problem:
```typescript
import { createStore, Dispatch } from 'redux';
import { Cmd, install, liftState, LoopReducer, loop, StoreCreator } from 'redux-loop';

interface ParentState {
    childState: ChildState;
}

interface ChildState {
    currentPage: number;
}

type ParentAction = { type: 'CHILD_ACTION', payload: { subAction: ChildAction } };

type ChildAction =
    { type: 'NEXT_CLICKED' } |
    { type: 'NEXT_PAGE_ANIMATION_FINISHED' };

const parentReducer: LoopReducer<ParentState, ParentAction> = (state, action) => {
    if (!state) {
        const [ childState ] = liftState(childReducer(undefined, null as any));
        return { childState };
    }

    console.log('parentReducer: processing action:\n', action);

    if (action.type === 'CHILD_ACTION') {
        const [ newChildState, cmd ] = liftState(childReducer(state.childState, action.payload.subAction));
        return loop(
            { ...state, childState: newChildState },
            Cmd.map(cmd, subAction => ({ type: 'CHILD_ACTION', payload: { subAction } })),
        );
    } else {
        console.log('parentReducer: unknown action:', action);
        return state;
    }
};

function scheduleNextPage(dispatch: Dispatch<ChildAction>) {
    setTimeout(() => dispatch({ type: 'NEXT_PAGE_ANIMATION_FINISHED' }), 1000);
}

const childReducer: LoopReducer<ChildState, ChildAction> = (state, action) => {
    if (!state) {
        return { currentPage: 0 };
    }

    if (action.type === 'NEXT_CLICKED') {
        return loop(
            state,
            Cmd.run(scheduleNextPage, { args: [Cmd.dispatch] }),
        );
    } else if (action.type === 'NEXT_PAGE_ANIMATION_FINISHED') {
        return {
            ...state,
            currentPage: state.currentPage + 1,
        };
    } else {
        console.log('childReducer: unknown action type:', action);
        return state;
    }
};

function main() {
    const store = (createStore as StoreCreator)(
        parentReducer,
        undefined,
        install(),
    );

    console.log('Main: simulating NEXT_CLICKED');
    store.dispatch({
        type: 'CHILD_ACTION',
        payload: {
            subAction: { type: 'NEXT_CLICKED' },
        },
    });

    setTimeout(() => {
        console.log('Main: final state:', store.getState());
    }, 2000);
}

main();
```

Compile and run with:
```
tsc test.ts --target es6 --module commonjs && node test.js
```
Output with redux-loop@master:
```
Main: simulating NEXT_CLICKED
parentReducer: processing action:
 { type: 'NEXT_CLICKED',
  payload: { subAction: { type: 'NEXT_CLICKED' } } }
parentReducer: processing action:
 { type: 'NEXT_PAGE_ANIMATION_FINISHED' }
parentReducer: unknown action: { type: 'NEXT_PAGE_ANIMATION_FINISHED' }
Main: final state: { childState: { currentPage: 0 } }
```

Output after fix:
```
Main: simulating NEXT_CLICKED
parentReducer: processing action:
 { type: 'CHILD_ACTION',
  payload: { subAction: { type: 'NEXT_CLICKED' } } }
parentReducer: processing action:
 { type: 'CHILD_ACTION',
  payload: { subAction: { type: 'NEXT_PAGE_ANIMATION_FINISHED' } } }
Main: final state: { childState: { currentPage: 1 } }
```

### DIFF
--- a/src/cmd.js
+++ b/src/cmd.js
@@ -160,7 +160,7 @@ export function executeCmd(cmd, dispatch, getState, loopConfig = {}) {
     case cmdTypes.MAP: {
       const possiblePromise = executeCmd(
         cmd.nestedCmd,
-        dispatch,
+        action => dispatch(cmd.tagger(...cmd.args, action)),
         getState,
         loopConfig
       );
--- a/test/cmd.spec.js
+++ b/test/cmd.spec.js
@@ -521,6 +521,28 @@ describe('Cmds', () => {
           { ...actionCreator2(action2), arg1, arg2 }
         ]);
       });
+
+      function dispatchAction(action) {
+        return Cmd.run(
+          (dispatch) => dispatch(action),
+          {
+            args: [Cmd.dispatch],
+          }
+        );
+      }
+
+      it('runs the actions from async dispatch through the tagger function', () => {
+        let action1 = actionCreator1(123);
+        let subCommand = dispatchAction(action1);
+        let arg1 = 'arg1',
+          arg2 = 'arg2';
+        let cmd = Cmd.map(subCommand, argTagger, arg1, arg2);
+        let result = executeCmd(cmd, dispatch, getState);
+        expect(result).toEqual(null);
+        expect(dispatch).toHaveBeenCalledWith(
+          { ...actionCreator2(action1), arg1, arg2 },
+        );
+      });
     });
 
     describe('Cmd.none', () => {
--- a/test/cmd.spec.js
+++ b/test/cmd.spec.js
@@ -523,12 +523,9 @@ describe('Cmds', () => {
       });
 
       function dispatchAction(action) {
-        return Cmd.run(
-          (dispatch) => dispatch(action),
-          {
-            args: [Cmd.dispatch],
-          }
-        );
+        return Cmd.run(dispatch => dispatch(action), {
+          args: [Cmd.dispatch]
+        });
       }
 
       it('runs the actions from async dispatch through the tagger function', () => {
@@ -539,9 +536,11 @@ describe('Cmds', () => {
         let cmd = Cmd.map(subCommand, argTagger, arg1, arg2);
         let result = executeCmd(cmd, dispatch, getState);
         expect(result).toEqual(null);
-        expect(dispatch).toHaveBeenCalledWith(
-          { ...actionCreator2(action1), arg1, arg2 },
-        );
+        expect(dispatch).toHaveBeenCalledWith({
+          ...actionCreator2(action1),
+          arg1,
+          arg2
+        });
       });
     });
 


bdwain:
I don't think the behavior you're implementing is desired. Actions that are dispatched through Cmd.dispatch should not be wrapped. They are not necessarily related to the reducer that is dispatching them. You should be able to dispatch something generic like a logout action from those methods and it should work. Wrapping it in the tagger would break that.

If you want actions to be wrapped, they will need to be done through the return values of commands passed to Cmd.map like normal, not Cmd.dispatch. Cmd.dispatch is just an escape hatch to do arbitrary dispatches from a Cmd if needed.

laszlopandy:
@bdwain 
We try to avoid using `Cmd.dispatch` in commands but sometimes it is unavoidable, for example if you need a command to attach an event listener or you need a command to return more than one action.
Here is our use case. We have an edit page which allows you to update the title & description of a document. This edit page is used in two ways:
1. on it's own (as a standalone app), as well as
2. in the create flow (as part of a larger app). 
We are following the elm architecture in building these subcomponents, and if `Cmd.dispatch` cannot be used with `Cmd.map` then it is much harder to make reuseable components which can be used in multiple apps.

laszlopandy:
We have identified a similar issue with commands that don't use `dispatch`.

Expanding on my previous `test.ts` example, the only changes here are:
 - the new `scheduleNextPagePromise` function and 
 - returning `Cmd.list([ ... ])` in the `childReducer`
```typescript
function scheduleNextPagePromise(): Promise<ChildAction> {
    return new Promise<ChildAction>(resolve => {
        setTimeout(() => resolve({ type: 'NEXT_PAGE_ANIMATION_FINISHED' }), 1000);
    });
}

const childReducer: LoopReducer<ChildState, ChildAction> = (state, action) => {
    if (!state) {
        return { currentPage: 0 };
    }

    if (action.type === 'NEXT_CLICKED') {
        return loop(
            state,
            Cmd.list([
                Cmd.run(scheduleNextPagePromise, { successActionCreator: a => a }),
            ]),
        );
    } else if (action.type === 'NEXT_PAGE_ANIMATION_FINISHED') {
        return {
            ...state,
            currentPage: state.currentPage + 1,
        };
    } else {
        console.log('childReducer: unknown action type:', action);
        return state;
    }
};
```

The expected and actual results are the same as above in the original description: if I use `Cmd.list` in `childReducer` the action is not wrapped by `Cmd.map` and instead is sent directly to the `parentReducer` (which prints `parentReducer: unknown action: ...`)

The issue here is that `Cmd.list` uses `dispatch()` internally to dispatch the action when the promise resolves. If you pass `{ batch: true }` then the issue goes away because `Cmd.list` returns the `Promise.all()` instead of dispatching each action as soon as it is ready.

This fix proposed in this PR also solves this issue as well as the original one.

bdwain:
Cmd.dispatch is not part of the elm architecture though. It's an escape hatch that's there for convenience. It's meant to dispatch arbitrary actions that are not related to the original action. Having it behave differently for Cmd.map would cause confusion and also break things for everyone who is relying on the current behavior.

If you want an action to be wrapped, just don't use Cmd.dispatch. It's there for convenience, but never NEEDED. If you need to dispatch multiple actions in response to something, just do it from the reducer with Cmd.action.

```js
//child reducer
case foo:
  return Cmd.map(myFunc, {
    successActionCreator: myFuncSuccess
  });
case myFuncSuccess:
  return Cmd.list([Cmd.action(action1()), Cmd.action(action2())]);
```
You could also pass the tagger function to the Cmd somehow to dispatch directly from it, though you'll need to come up with a way to pass the tagger down to the child reducer.

For the Cmd.list issue, you can use the batch option like you mentioned. That's a known issue and [mentioned in the docs](https://redux-loop.js.org/docs/api-docs/cmds.html#cmdnone). Though I'm thinking we could maybe address it by doing something similar to what you're doing here, and then that issue would go away. But the behavior of Cmd.dispatch definitely needs to stay the same.

laszlopandy:
To clarify, when I said we need to dispatch multiple actions from one command I means different actions happening at different times. For example, callbacks from an event. I couldn't find in the docs the correct way to get events without `Cmd.dispatch`. Elm has the subscriptions concept which doesn't seem to be present in redux-loop. You can put these in `componentDidMount` but that doesn't seem right for events which are not UI related.

```typescript
case foo:
    return Cmd.map(
        dispatch => gameEngine.addEventListener('modeChanged', e => dispatch(actions.modeChanged(e.newMode))),
        {
            args: [Cmd.dispatch],
        }
);
```

bdwain:
Hm i see. this is an interesting problem. 

Vanilla redux does have [subscriptions](https://redux.js.org/api/store#subscribelistener),  though I'm not sure it's the same thing you're looking for?

In your example, did you mean to use Cmd.run, not Cmd.map? Can you pass the tagger  down to the nested reducer so that you can wrap the action that you pass to dispatch?

laszlopandy:
Yes I meant Cmd.run, sorry. 
We could pass the tagger down, and probably we will do this. We can work around it of course.

I'm more concerned in principle that this is a gotcha in the library. Several of my colleagues have hit the same issue (either with Cmd.dispatch or Cmd.list) and spent hours in the debugger trying to figure out why their actions aren't getting to the reducer -- until I talk to them and figure out it's the same issue coming up again.

Secondly, you said that Cmd.dispatch is a convenience but never needed. So what is the recommended way to use addEventListener? Even ignoring the tagger issue, from the root of my app, I haven't seen in the docs the recommended way to attach to external events. Thanks.

bdwain:
I see. yes Cmd.dispatch is required for now if you want to dispatch a variable number of actions or dispatch them at different times. I think the current Cmd.dispatch behavior needs to stay the same, but there are a few options.

1. Improve documentation to better explain the gotcha. Since Cmd.map is not used that much, I think this could maybe be sufficient.
2. Maybe a new Cmd type could help here? Not sure what it would look like.
3. Add a new symbol, like `Cmd.map.dispatch`, which does the wrapping like you do here.

thoughts? I'm leaning 1 for now. If there was an idea that would work for 2, I'd be open to it, but I don't have one. 3 would kind of solve the problem, but is still kind of a gotcha, and I'm worried it could make the API more confusing.

Thoughts?

laszlopandy:
Yes I do have some thoughts on this. While I organize them, is this the best place to discuss this or is there a better forum for discussing API changes?

laszlopandy:
As the first priority, we should fix the `Cmd.list` gotcha so wrapped actions from sub-reducers don't have to use `{ batch: true }` as a work around. `Cmd.list` is not an escape hatch, and there is a clear expected behaviour.

To deal with the `Cmd.dispatch` confusion, what if we passed the store to the command function instead of separate `getState` and `dispatch` arguments?
```typescript
case foo:
    return Cmd.runWithStore(
        store => gameEngine.addEventListener('modeChanged', e => store.dispatch(e.newMode)),
        {
            successActionCreator: actions.modeChanged,
        }
); 
```
In this example, seeing `store.dispatch(e.newMode)` makes it clear to me that it will go directly to the store (ignoring any `Cmd.map` that has been applied). This might fix the confusion without having to read the docs.

I would also get rid of the symbols `Cmd.getState` and `Cmd.dispatch` and instead split `Cmd.run` into two:
 * `Cmd.run` - simply calls `func(...args)`
 * `Cmd.runWithStore` - calls `func(store, ...args)`

In addition to making the API more explicit, this would make it possible to write complete type definitions for the command API. Currently it is not possible to type-check usages of `Cmd.getState` and `Cmd.dispatch` because of their dynamic position in the list of arguments. 

Here is what the .d.ts would look like:
```typescript
export function run<A extends Action, B extends Action, Args extends Array<any>, Value>(
  f: (...args: Args) => Value,
  options?: {
    args: Args;
    failActionCreator?: (value: Value) => A;
    successActionCreator?: (err: string) => B;
    forceSync?: boolean;
    testInvariants?: boolean;
  }
): RunCmd<A>;

export function runWithStore<A extends Action, B extends Action, Args extends Array<any>, Value>(
  f: (store: Store, ...args: Args) => Value,
  options?: {
    args: Args;
    failActionCreator?: (value: Value) => A;
    successActionCreator?: (err: string) => B;
    forceSync?: boolean;
    testInvariants?: boolean;
  }
): RunCmd<A>;
```

bdwain:
Ok I have an idea of how we can fix the gotcha for the list Cmd type. I'll get a PR set up for that. It's a little weird in implementation but it's invisible to the user (except the removal of the gotcha) so I think it's good to do. It may be a breaking change, which is not ideal, but that's fine.

As for adding a `runWithStore` Cmd type, that's a big API change (it would require fixes everywhere for people who use the symbols now) and I'm not sure  it's worth it when all it really does is add some clarity on what is doing the dispatching. I also don't love dictating the parameter orderings. You made #233 to discuss that though so we can talk more about it there. 

I think improved documentation on Cmd.dispatch can add clarity about who is dispatching without requiring everyone to change existing apps. I'm not so concerned with type safety since #222 addresses that anyway.

Cmd.map is not used so much that I want to make major changes to the API for it. I know it's not as convenient to have to pass the tagger function down, but for an escape hatch like Cmd.dispatch, I don't think it's the end of the world.

I'm still open to other ways to simplify this because I see how it could be annoying, but I don't want to make breaking changes to the API unless it adds more significant improvements to other Cmd types too. But maybe there are ways to do this that don't break existing APIs?

bdwain:
Created #235 to fix the list issue.

PR #222, opened 2019-08-09
Make Cmd.run typesafe

domarmstrong:
I notice there is a stale PR that tackles the same problem.

This PR goes a step further with argument inference while still allowing the special symbols to be used if the fn passed has arguments that would match the types of Dispatch or GetState (although I can't distinguish between them as the type symbol === symbol)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -57,14 +57,14 @@ export interface MapCmd<A extends Action> {
   simulate(simulations?: CmdSimulation | MultiCmdSimulation): A[] | A | null
 }
 
-export interface RunCmd<A extends Action> {
+export interface RunCmd<SuccessAction extends Action, FailAction extends Action = Action> {
   readonly type: 'RUN';
   readonly func: Function;
   readonly args?: any[];
-  readonly failActionCreator?: ActionCreator<A>;
-  readonly successActionCreator?: ActionCreator<A>;
+  readonly failActionCreator?: ActionCreator<FailAction>;
+  readonly successActionCreator?: ActionCreator<SuccessAction>;
   readonly forceSync?: boolean;
-  simulate(simulation: CmdSimulation): A
+  simulate(simulation: CmdSimulation): SuccessAction | FailAction;
 }
 
 //deprecated types
@@ -96,6 +96,9 @@ export namespace Cmd {
   export const dispatch: unique symbol;
   export const getState: unique symbol;
   export const none: NoneCmd;
+  export type Dispatch = <A extends Action>(a: A) => void;
+  export type GetState = <S>() => S;
+
   export function action<A extends Action>(action: A): ActionCmd<A>;
   export function batch<A extends Action>(cmds: CmdType<A>[]): BatchCmd<A>;
   export function sequence<A extends Action>(cmds: CmdType<A>[]): SequenceCmd<A>;
@@ -115,16 +118,30 @@ export namespace Cmd {
     args?: any[]
   ): MapCmd<A>;
 
-  export function run<A extends Action, B extends Action>(
-    f: Function,
+  // Allow the use of special dispatch | getState symbols
+  type ArgOrSymbol<T> = {
+    [K in keyof T]: T[K] extends GetState
+      ? typeof getState
+      : T[K] extends Dispatch
+        ? typeof dispatch
+        : T[K];
+  }
+  export type PromiseResult<T> = T extends Promise<infer U> ? U : T;
+
+  export function run<
+    Func extends (...args: any[]) => Promise<any> | any,
+    SuccessAction extends Action,
+    FailAction extends Action,
+    >(
+    f: Func,
     options?: {
-      args?: any[];
-      failActionCreator?: ActionCreator<A>;
-      successActionCreator?: ActionCreator<B>;
+      args?: ArgOrSymbol<Parameters<Func>>;
+      failActionCreator?: (error: any) => FailAction;
+      successActionCreator?: (value: PromiseResult<ReturnType<Func>>) => SuccessAction;
       forceSync?: boolean;
       testInvariants?: boolean;
-    }
-  ): RunCmd<A>;
+    },
+  ): RunCmd<SuccessAction, FailAction>;
 }
 
 export type ReducerMapObject<S, A extends Action = AnyAction> = {
--- a/test/typescript/loopReducer.ts
+++ b/test/typescript/loopReducer.ts
@@ -47,7 +47,7 @@ interface IFetchFooSuccess {
   type: typeof FETCH_FOO_SUCCESS
 }
 
-const fetchFooSuccess = (): IFetchFooSuccess => ({
+const fetchFooSuccess = (result: string): IFetchFooSuccess => ({
   type: FETCH_FOO_SUCCESS,
 })
 
@@ -55,7 +55,11 @@ interface IFetchFooFailure {
   type: typeof FETCH_FOO_FAILURE
 }
 
-const fetchFooFailure = (): IFetchFooFailure => ({
+class CustomError extends Error {
+  public prop = 'myprop';
+}
+
+const fetchFooFailure = (err: CustomError): IFetchFooFailure => ({
   type: FETCH_FOO_FAILURE,
 })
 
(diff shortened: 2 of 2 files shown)


index.d.ts
```diff
@@ -96,6 +96,9 @@ declare namespace Cmd {
   export const dispatch: symbol;
   export const getState: symbol;
   export const none: NoneCmd;
+  export type Dispatch = <A extends Action>(a: A) => void;
```
bdwain: dispatch technically returns a promise

bdwain:
Is any of this using a new typescript feature that requires a version we don't currently require, or is it a breaking change at all? I think we currently require at least 2.8, but I'm not too familiar with typescript as I don't use it myself.

index.d.ts
```diff
@@ -115,16 +118,30 @@ declare namespace Cmd {
     args?: any[]
   ): MapCmd<A>;
 
-  export function run<A extends Action, B extends Action>(
-    f: Function,
+  type InferArgs<T> = T extends (...args: infer T) => any ? T : never;
```
bdwain: what exactly does infer do here?

index.d.ts
domarmstrong: A promise of the action or void, ? `<A extends Action>(a: A) => Promise<A>`. The redux typing returns just the action:
```
export interface Dispatch<A extends Action = AnyAction> {
  <T extends A>(action: T): T
}
```

index.d.ts
domarmstrong: It basically says from this type this is the argument I want to infer my variable T to.

Its a conditional type that only matches the left hand size, its a bit of a quirky typescript syntax.

Simplified examples:
```
type NumberName<T> = T extends number ? 'number' : 'notNumber';
type N = NumberName<2>; // 'number' < literal
type S = NumberName<'a'>; // 'notNumber' < literal
```
Then add never:
```
type NumberName<T> = T extends number ? 'number' : never;
type N = NumberName<2>; // 'number' < literal
type S = NumberName<'a'>; // never;
```
Then add infer;
```
type Args<T> = T extends (...args: infer A) => any ? A : never;
type A = Args<(a: number, b?: string) => void>; // [number, (string | undefined)]
```

Hopefully that makes sense?

domarmstrong:
> Is any of this using a new typescript feature that requires a version we don't currently require, or is it a breaking change at all? I think we currently require at least 2.8, but I'm not too familiar with typescript as I don't use it myself.

Yes it does require ^3.0.1 because of the generic rest parameter https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#generic-rest-parameters. I don't think its possible to infer the argument before that, but it is one of the more useful bits.

index.d.ts
laszlopandy: Is this not the same as TypeScript's built-in `Parameters`?
See lib.es5.d.ts:
```
/**
 * Obtain the parameters of a function type in a tuple
 */
type Parameters<T extends (...args: any) => any> = T extends (...args: infer P) => any ? P : never;
```

index.d.ts
```diff
@@ -115,16 +118,30 @@ declare namespace Cmd {
...
+  // Allow the use of special dispatch | getState symbols
+  type ArgOrSymbol<T> = {
+    [K in keyof T]: T[K] extends Dispatch | GetState
+      ? typeof dispatch | typeof getState | T[K]
+      : T[K];
+  }
```
laszlopandy: To be honest I didn't think that the type system could express `Cmd.run` API was fully. I'm glad to learn that TypeScript is this powerful, but if we have to resort to this much magic in the type-definitions, maybe we should revisit if this API can be simplified first.

bdwain:
@laszlopandy can you review this? Since it requires typescript v3, it sounds like it'll be a breaking change to introduce. How widespread is typescript v3 usage? Is it a difficult upgrade from v2?

lorefnon:
TS Upgrades are generally smooth, and v3 was released several months ago, so I don't anticipate updating to TS v3 to be a major issue for people. We have been updating to every minor version sequentially in <1 day over the last two years.

But if the API significantly changes to make it easier to typecheck that can be a larger breaking change and a hurdle towards upgrading.

I think it would be good if this change is merged and any API changes (if needed) be taken up in separate breaking changes.

bdwain:
@lorefnon awesome. Then I think we can just do a v6 to release this. There is another breaking change I want to release anyway

index.d.ts
domarmstrong: Yes, thanks. I've updated to use Parameters

index.d.ts
nweber-gh: Why do we define and export `Dispatch` in here at all? Let's just import it from Redux.

index.d.ts
```diff
@@ -115,16 +118,29 @@ export namespace Cmd {
...
+    SuccessAction extends Action,
+    FailAction extends Action,
+    >(
+    f: Func,
     options?: {
-      args?: any[];
-      failActionCreator?: ActionCreator<A>;
-      successActionCreator?: ActionCreator<B>;
+      args?: ArgOrSymbol<Parameters<Func>>;
+      failActionCreator?: (error: Err) => FailAction;
```
laszlopandy, line 138: A promise can fail with `error: any`. It is not always an `Error`. To be honest, I would leave this as `any`. In TypeScript Promises cannot specify the error type, so it might not be worth it to specify an error type here.

laszlopandy:
@domarmstrong
> (although I can't distinguish between them as the type symbol === symbol)

Is this still true? Does this mean if you switch the arguments accidentally it doesn't give a type error?

index.d.ts
domarmstrong: I've updated this, it seems sensible

domarmstrong:
> @domarmstrong
> 
> > (although I can't distinguish between them as the type symbol === symbol)
> 
> Is this still true? Does this mean if you switch the arguments accidentally it doesn't give a type error?

No, I've fixed this. It should correctly error now if you use the wrong symbol, which I've tested in the loopReducer.ts file.

```diff
-    [K in keyof T]: T[K] extends Dispatch | GetState
-      ? typeof dispatch | typeof getState | T[K]
-      : T[K];
+    [K in keyof T]: T[K] extends GetState
+      ? typeof getState
+      : T[K] extends Dispatch
+        ? typeof dispatch
+        : T[K];
```

laszlopandy:
Thanks @domarmstrong in that case I suggest we merge this now and deal with any new issues when we prepare for the next release. Thanks for your work on this.

(FYI: it's not necessary to force push to keep a single clean commit; we squash merge PRs in this repo).

index.d.ts
bdwain: @domarmstrong @laszlopandy any thoughts on this? is the reason we don't just reuse the redux dispatch type because the return type is different?

and @domarmstrong i think i forgot to respond to your original comment months ago.  it actually returns a promise with the new state object.

index.d.ts
```diff
@@ -115,16 +118,30 @@ export namespace Cmd {
...
+    Func extends (...args: any[]) => Promise<any> | any,
+    SuccessAction extends Action,
+    FailAction extends Action,
+    >(
+    f: Func,
     options?: {
-      args?: any[];
-      failActionCreator?: ActionCreator<A>;
-      successActionCreator?: ActionCreator<B>;
+      args?: ArgOrSymbol<Parameters<Func>>;
```
nweber-gh, line 138: I wonder if we could use `infer` to actually determine which arguments need to be provided to the function so that we can get type checking.

index.d.ts
laszlopandy: I think we should have a different type because the Redux type definitions show that dispatch returns an action, but in redux-loop we return a `Promise<Action>`. I think this warrants making a new type, but yes we should be consistent about this. 
Also I never understood why dispatch would return anything in redux. If someone knows, I would be glad to understand it.

index.d.ts
laszlopandy: Can you explain what you mean? `Parameters` is a typescript built-in which uses infer internally.

nweber-gh: Ah, I didn't realize that. I think this is already doing what I was hoping it was, haha. Disregard.

PR #228, opened 2019-10-30
Fix typescript declaration style

aeirola:
The current typescript declarations seem to be a strange mix of [global declarations](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-d-ts.html) and [module declarations](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html). As the package is intended to be used as a module, the appropriate declaration style is to use the `export` keyword when defining the elements of the module.

This is similar to the style for example the react-redux package [types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-redux/index.d.ts) uses.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -85,14 +85,14 @@ export interface LoopConfig {
   readonly DONT_LOG_ERRORS_ON_HANDLED_FAILURES: boolean;
 }
 
-declare function install<S>(config?: LoopConfig): StoreEnhancer<S>;
+export function install<S>(config?: LoopConfig): StoreEnhancer<S>;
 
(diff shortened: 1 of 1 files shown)


bdwain:
thanks! Will this make a difference for current users of the types? Would it be a breaking change?

aeirola:
> thanks! Will this make a difference for current users of the types? Would it be a breaking change?

No, shouldn't affect existing users unless they import the library in some really strange and undocumented way.

bdwain:
awesome i'll merge it now then. thanks!

PR #232, opened 2020-02-24
Cmd.setTimeout, Cmd.setInterval

laszlopandy:
Add a new type of action which schedules another command to be dispatched using `setTimeout()` or `setInterval()`. This is a very useful feature for starting/stopping animations and scheduling timeouts. 

Having an explicit action type makes the intention of the code clear -- you don't need to use `Cmd.run()` for something which is not an external side-effect.


bdwain:
I like the thought behind this, but I'd like to see it be more generic. I think it would be good to build in support for intervals as well as timeouts, and also to support cancellation. Without those, I wouldn't feel comfortable adding it to the API because a timeout that can't be cancelled is going to shoot a lot of people in the foot (and you can do this pretty easily with Cmd.run).

I also think it would be good to see if we could just make all effects delayable, rather than make individual ones for action.

Here's an API idea I had.

`Cmd.delay(otherCmd, {
	delay: 1000,
	interval: true,
	onScheduledActionCreator: myActionCreator
})`

and then you could use the scheduled action creator to store the timeout id in case you wanted to cancel it later.

Thoughts?

laszlopandy:
Great ideas to make this more general. Scheduling intervals is something we have in our code, and it always feels more awkward that it should be. 

My original thought was that you can use `Cmd.delay()` to do intervals by returning another `Cmd.delay()` after each action is processed. Similarly, there is no need to cancel because you can just set a flag in your state that the next action should be ignored. While this works in theory it isn't very natural especially for programmers who are used to using `setTimeout()` and `setInterval()`. Which begs the question: should we just use those names (even though I think `delay` is much better) because of familiarity?

`Cmd.setTimeout(otherCmd, 1000, Actions.setTimeoutHandle)`
`Cmd.setInterval(otherCmd, 1000, Actions.setIntervalHandle)`

laszlopandy:
@bdwain I implemented `Cmd.schedule` and `Cmd.scheduleRepeating`, and updated the tests. Let me know what you think, and if it's good I'll update the docs too.

bdwain:
Sorry for the delay in looking at this. With everything going on in the world I lost track of things.

I think I'm torn between the name delay and setTimeout/setInterval. But i definitely like them both more than schedule. I'm also torn between having 2 separate Cmd types or one with a boolean option to switch between the two. I feel like there's cognitive overhead when you add extra Cmd types. Parallel and non-parallel lists used to be two different command types, but then we combined them into one and I think it worked better (See discussion [here](https://github.com/redux-loop/redux-loop/pull/151#issuecomment-332058787))

I think we should either do Cmd.delay with a boolean switch option or setTimeout and setInterval as separate ones (sharing most of their implementation) because setTimeout and setInterval aren't new concepts to people so there should be little cognitive overheard, but delay would. I'm pretty torn though between the two though. Thoughts? I think I'm leaning towards separate Cmds, just because it's consistent with web APIs so it'd be pretty straightforward.

Other than that, the API seems good to me, except I'd like to have the third parameter be an options object. I've had that come back to bite me too many times. We'll want to add some option in the future but having a never ending parameter list is hard to use. 

I'll comment on implementation details, but I like the  direction this is heading.

src/cmd.js
```diff
@@ -258,6 +288,46 @@ function action(actionToDispatch) {
...
 
+function schedule(nestedCmd, delayMs, onScheduledActionCreator) {
+  return _schedule(nestedCmd, delayMs, onScheduledActionCreator, false);
+}
+
+function scheduleRepeating(nestedCmd, delayMs, onScheduledActionCreator) {
+  return _schedule(nestedCmd, delayMs, onScheduledActionCreator, true);
+}
+
+function _schedule(nestedCmd, delayMs, onScheduledActionCreator, isRepeating) {
```
bdwain: since this isn't a publicly accessible function I think the _ name is kind of unnecessary. Just call it whateverHelper or something (depending on the final name of the Cmd(s))

src/cmd.js
```diff
@@ -258,6 +288,46 @@ function action(actionToDispatch) {
...
+    nestedCmd,
+    delayMs,
+    isRepeating,
+    onScheduledActionCreator,
+    simulate: simulateSchedule
+  });
+}
+
+function simulateSchedule(simulation) {
+  return this.nestedCmd.simulate(simulation);
```
bdwain: hmmm we should do more than simulate the nested command. We want to be able to test the action creator too. 

i think the passed in simulation object should contain something like interval/timeout id, as  well as a nested cmd simulation object, and then the result should contain the nested result and the action that would get dispatched with the id.

something like this

```js
  return {
     nestedResult: this.nestedCmd.simulate(simulation.nestedSimulation),
     action: this.onScheduledActionCreator(simulation.timeoutId)
  };
```

src/cmd.js
```diff
@@ -144,6 +145,32 @@ function handleSequenceList(
...
+      dispatch,
+      getState,
+      loopConfig
+    );
+    if (cmdPromise) {
+      cmdPromise.then(actions => {
+        if (!actions.length) {
+          return;
+        }
+        return Promise.all(actions.map(dispatch));
```
bdwain: there's no need to return here. nothing is looking  at the result of the promise. just say `actions.map(dispatch)`.

bdwain: actually, map and forEach pass extra params, which we should not do. so i would say `actions.forEach(action => dispatch(action))`

bdwain:
I polled everyone I work with and they all prefer setTimeout/setInterval also. I think that is a good way to name it

laszlopandy:
Hey @bdwain. I'm also sorry for the delay. I lost some github notifications and didn't check up on this.

I also agree that `Cmd.setTimeout` and `Cmd.setInterval` is preferable because it matches the JS api and everyone knows what they mean. On the other hand that means we should inherit the cancel names too.

Are you okay with these names?
 * Create methods: `Cmd.setTimeout()` and `Cmd.setInterval()`
 * Cancel methods: `Cmd.clearTimeout()` and `Cmd.clearInterval()`
 * Types in Typescript and flow: `SetTimeoutCmd<A>` and `SetIntervalCmd<A>` with `type: "SET_TIMEOUT"` and `type: "SET_INTERVAL"` respectively.

The other option is to have a single type `DelayedCmd<A>` (`type: "DELAYED"`) with an `isRepeating` attribute and make it clear that `Cmd.setTimeout()` and `Cmd.setInterval()` are just two different constructors for the same kind of command. WDYT?

src/cmd.js
laszlopandy: Simulate has to return `Array<Action>` or `Action` or `null`, otherwise it will break if nested inside another command like Cmd.list. Probably we should simplify this and have all simulate methods return an `Array<Action>` but this would be a breaking change so let's leave it for later.

src/cmd.js
laszlopandy: Yep. Thanks, I missed this one.

This brings up a more important issue though. By using `dispatch()` internally (instead of returning promises) `Cmd.setTimeout` and `Cmd.setInterval` will have the same gotcha as `Cmd.list`. When wrapping these commands in `Cmd.map`, the tagger is not called and sub-actions are dispatched directly to the root reducer. I would like to fix that gotcha before we release this new feature.

bdwain: let's merge my other pr for that issue first, and then you can rebase this pr and address the issue. You had one comment I needed clarification on, but once you're good with it I'll merge the PR

bdwain:
ah yea I guess we do want clearTimeout and clearInterval cmds also. As far as separate or shared types, I'll defer to you on that one since i haven't done a ton of typescript yet. Though I'm curious what @nweber-gh thinks.

src/cmd.js
```diff
@@ -258,6 +291,71 @@ function action(actionToDispatch) {
   });
 }
 
+function clearTimeoutCmd(timerId) {
+  return run(clearTimeout, { args: [timerId] });
```
bdwain, line 295: this will actually cause a weird error `TypeError: Illegal invocation`

I don't totally understand why, but it has to do with the context of the window object when you call the function.

The way to get around it is to make a helper function and call it in there. Here is what I have in a project of mine.

```js
export function clearDelayedEffect(isTimeout, id) {
  const method = isTimeout ? clearTimeout : clearInterval;
  method(id);
}
```

and then you can just return here

```js
return run(clearDelayedEffect, {args: [false, timerId]})
```

you could also do 2 separate helpers to avoid the extra param. doesn't really matter

src/cmd.js
```diff
@@ -258,6 +291,71 @@ function action(actionToDispatch) {
...
+  return delay(nestedCmd, delayMs, options, false);
+}
+
+function setIntervalCmd(nestedCmd, delayMs, options = {}) {
+  return delay(nestedCmd, delayMs, options, true);
+}
+
+function delay(nestedCmd, delayMs, options, isRepeating) {
+  if (process.env.NODE_ENV !== 'production') {
+    const name = isRepeating ? 'Cmd.setInterval' : 'Cmd.setTimeout';
```
bdwain, line 312: i feel like if we need the name here, we should pass it in. If we rename the cmd, we shouldn't have to look here to change that.

src/cmd.js
bdwain: Now is the time to do breaking changes. The next release is going to have breaking changes in it anyway. And the action creator is the main thing simulate is designed to test, so I think it'd be good to include it.

That being said, we can probably avoid the breaking changes anyway. I didn't realize it would break the list and map simulations. What if we returned an array instead. 

```js
return this.nestedCmd.simulate(simulation.nestedSimulation).concat(this.onScheduledActionCreator(simulation.timeoutId))
```

bdwain: wait i see you already did that. perfect

test/cmd.spec.js
```diff
@@ -227,6 +227,27 @@ describe('Cmds', () => {
       });
     });
 
+    describe('Cmd.setTimeout', () => {
```
bdwain, line 230: can you add tests for cmd.setinterval?

bdwain:
sorry didn't mean to close this. 

the code looks good. Besides my few comments, and the conflicts that got introduced from #235, I like how this looks.

bdwain:
also, if you end up merging before I look at this again, please remember to squash before you merge

bdwain:
also, don't forget to use the wrappedDispatch function as discussed in https://github.com/redux-loop/redux-loop/pull/235/files#r397694600

src/cmd.js
bdwain: my pr is merged

bdwain:
@laszlopandy do you think you'll get back to this soon? I'd love to release a new version soon

index.d.ts
```diff
@@ -49,6 +49,15 @@ export interface ActionCmd<A extends Action> {
   simulate(): A;
 }
 
+export interface DelayCmd<A extends Action> {
```
bdwain, line 52: i wouldn't  worry about flow. i think it makes sense to drop support for it

laszlopandy: Okay great. I will remove flow changes.

laszlopandy:
@bdwain Yep. I was busy with family stuff recently but I'm been thinking about how to finish this up. I'll set aside some time this week.

index.d.ts
bdwain: sounds good!

bdwain:
sounds good thanks @laszlopandy

laszlopandy:
TODO:
 - [ ] Fix conflicts
 - [ ] Remove flow definition changes
 - [ ] Use two separate interfaces in TypeScript definitions
 - [ ] Use wrappedDispatch
 - [ ] Add tests for `Cmd.setInterval`
 - [ ] Verify that `TypeError: Illegal invocation` is fixed by https://github.com/redux-loop/redux-loop/pull/242/

src/cmd.js
laszlopandy: This PR is using `wrappedDispatch` now.

src/cmd.js
laszlopandy: The `.apply()` in https://github.com/redux-loop/redux-loop/pull/242/ fixed the `TypeError: Illegal invocation`.

src/cmd.js
laszlopandy: The name is passed in now.

index.d.ts
laszlopandy: Flow changes removed.

docs/api-docs/cmds.md
```diff
@@ -357,6 +389,16 @@ function reducer(state = { /* ... */ }, action) {
 }
 ```
 
+## Convenience helpers
```
bdwain, line 392: i wouldn't call these convenience helpers. They're regular commands just like setTimeout and clearTimeout. I'd rather just throw them next to setTimeout and setInterval above.

docs/api-docs/cmds.md
```diff
@@ -357,6 +389,16 @@ function reducer(state = { /* ... */ }, action) {
...
+
+### `Cmd.clearTimeout(timerId)`
+Cancel's a delayed command execution created by `Cmd.setTimeout()`.
+Equivalent to `Cmd.run(clearTimeout, { args: [timerId] })`.
```
bdwain, line 396: I don't think this is true. Because of issues with the `this` param and how clearTimeout is called. I've needed to create a wrapper function that calls `clearTimeout` rather than passing it right to Cmd.run

laszlopandy: The TypeError you mentioned is fixed by #242. You can now pass `createTimeout` directly to `Cmd.run()` and it works as expected.

bdwain: ah forgot about that. nvm then

index.d.ts
```diff
@@ -61,6 +61,22 @@ export interface ActionCmd<A extends Action> {
...
 
+export interface SetTimeoutCmd<A extends Action = never> {
+  readonly type: 'SET_TIMEOUT';
+  readonly nestedCmd: CmdType;
+  readonly delayMs: number;
+  readonly scheduledActionCreator?: (timerId: number) => A;
+  simulate(timerId: number, nestedSimulation?: CmdSimulation | MultiCmdSimulation): A[] | A | null
+}
+
+export interface SetIntervalCmd<A extends Action = never> {
```
nweber-gh, line 72: Do we support passing arguments, similar to `RunCmd`, for the delayed commands? Both `setTimeout` and `setInterval` support an arbitrary set of args, which get sent to the callback function, so I think it would make sense to support `args` in these commands as well.

example
```
setTimeout((message, count) => console.log(message, count), 500, 'hello world', 7);
```

nweber-gh: Or.. Perhaps that doesn't make sense, since this is delaying other commands. We could just delay a `RunCmd` to get args. It might make sense to support `args` to pass to the `scheduleActionCreator` however. What do you guys think?

bdwain: Yea I think the callback args aren't really necessary. Since this is taking other Cmds, we can just pass those args directly to the cmd if needed. And for passing args to the action creator, we can just use arrow functions like we've done with Cmd.run.


```ts
const runCmd = Cmd.run(
  (message, count) => console.log(message, count),
  {args: ['hello world', 500], successActionCreator: whatever}
)
return loop(newState, Cmd.setTimeout(runCmd, 1234, {scheduledActionCreator: id => actionCreator(id, arg, arg2)}))
```

laszlopandy:
@bdwain This PR is done, feel free to merge it whenever you are ready.

test/cmd.spec.js
```diff
@@ -227,6 +227,48 @@ describe('Cmds', () => {
...
+        let action = actionCreator1(123);
+        let cmd = Cmd.setTimeout(Cmd.action(action), 100, {
+          scheduledActionCreator: actionCreator2
+        });
+        let result = executeCmd(cmd, dispatch, getState);
+
+        await expect(result).resolves.toEqual([
+          actionCreator2(expect.anything())
+        ]);
+        await dispatchPromise;
```
bdwain, line 245: shouldn't we mock the timers and wait for those rather than just resolve as soon as dispatch is called? That would allow us to verify we're actually waiting for the right amount of time, and would allow us to improve the setInterval test by verifying it actually runs multiple times

laszlopandy: Yes, do you have a nice way to do this, or would you just overwrite the global `setInterval` and restore it in an after block.

bdwain: jest can do this https://jestjs.io/docs/en/timer-mocks

laszlopandy: Done, please check it again.

bdwain: can you do the same thing for setTimeout? basically the same as the setInterval test, except we verify that we don't dispatch more actions after the time elapses again

laszlopandy: Done.

bdwain:
thanks @laszlopandy !

PR #235, opened 2020-03-25
Wrap actions from unbatched lists in the tagger from Cmd.map

bdwain:
@laszlopandy this should address the list issue from #218

### DIFF
--- a/src/cmd.js
+++ b/src/cmd.js
@@ -66,23 +66,25 @@ function handleRunCmd(cmd, dispatch, getState, loopConfig = {}) {
 function handleParallelList(
   { cmds, batch = false },
   dispatch,
+  wrappedDispatch,
   getState,
   loopConfig = {}
 ) {
   const promises = cmds
     .map(nestedCmd => {
-      const possiblePromise = executeCmd(
-        nestedCmd,
+      const possiblePromise = executeCmdInternal({
+        cmd: nestedCmd,
         dispatch,
+        wrappedDispatch,
         getState,
         loopConfig
-      );
+      });
       if (!possiblePromise || batch) {
         return possiblePromise;
       }
 
       return possiblePromise.then(result => {
-        return Promise.all(result.map(a => dispatch(a)));
+        return Promise.all(result.map(a => wrappedDispatch(a)));
       });
     })
     .filter(x => x);
@@ -101,6 +103,7 @@ function handleParallelList(
 function handleSequenceList(
   { cmds, batch = false },
   dispatch,
+  wrappedDispatch,
   getState,
   loopConfig = {}
 ) {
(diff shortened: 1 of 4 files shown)


src/cmd.js
```diff
@@ -82,7 +84,7 @@ function handleParallelList(
       }
 
       return possiblePromise.then(result => {
-        return Promise.all(result.map(a => dispatch(a)));
+        return Promise.all(result.map(a => wrappedDispatch(a)));
```
bdwain: the key is here. We use the wrapped dispatch rather than the vanilla one.

I could see something like this maybe being needed in the new timeout/delay Cmd types also, but we can deal with that after this is merged (or I will add it if that gets merged first)

src/cmd.js
```diff
@@ -145,6 +155,16 @@ function handleSequenceList(
 }
 
 export function executeCmd(cmd, dispatch, getState, loopConfig = {}) {
+  return executeCmdInternal(cmd, dispatch, dispatch, getState, loopConfig);
```
laszlopandy, line 158: Should we consider putting these four in an object at some point? It would be less error-prone than passing all four down.

src/cmd.js
laszlopandy: Yes, I also realized this will be an issue for Cmd.setTimeout. And yes we can solve it separately.

laszlopandy:
@bdwain I had another idea after reading the discussion of the return type of dispatch https://github.com/redux-loop/redux-loop/issues/237

What if instead of returning `Promise<Array<Action>>`, `executeCmd` returned `Array<Promise<Action>>`? In that way you could maintain the behaviour of parallel list execution (without batch) by returning a list of promises which resolve at different times. The `dispatch` would be used solely for run commands that request `Cmd.dispatch`.

The disadvantage is that this idea would require a bigger refactor. On the plus side it makes `executeCmd` conceptually simpler in that all actions would be returned. Whereas now some are returned (to be dispatched by the caller) and some are dispatched internally.

src/cmd.js
bdwain: By a context object do you just mean one object with properties `cmd`, `dispatch`, `wrappedDisaptch`, etc... ?

What kinds of errors do you see happening with individual params? I didn't think this was a problematic pattern

bdwain:
@laszlopandy can you make a separate issue to discuss that? It sounds interesting, though I'm not totally clear on what problem it's solving. But I think whatever it is would be separate from this change.

bdwain:
@laszlopandy one other thing is that that change sounds like it could potentially be a breaking change, and if so, we'll want to wait to do a v6 until we have it merged, since this and a few of the other changes going out are breaking too.

src/cmd.js
laszlopandy: An error could be accidentally passing dispatch in place of wrappedDispatch. In my experience it's harder to mix up named attributes than it is to accidentally swap position arguments.

I think if there were 10 parameters here we would use an object. For now it's just a question of where you draw the line. Either way is fine.

bdwain: that's a good point. The parameter list could grow, and the 2 dispatches are definitely easy to mix up. i'll switch it  to use an object

bdwain: executeCmd (and   the internal helper) now take an object

PR #238, opened 2020-04-20
executeCmd uses context object to avoid repetition

laszlopandy:
Use a context object for the (mostly) unchanging parameters used by `executeCmd` and its helper functions.

@bdwain This is more what I meant regarding passing the extra parameters in an object. `context` might be the wrong name, but I think this makes the code cleaner.


bdwain:
this is nicer i agree

PR #239, opened 2020-04-20
Simulate always return an array.

laszlopandy:
So we don't always have to check the result with Array.isArray() and check if it is null.

This will be a breaking change, so let's do it now. I think it makes the API more consistent.

### DIFF
--- a/docs/api-docs/cmds.md
+++ b/docs/api-docs/cmds.md
@@ -26,10 +26,10 @@ it doesn't cause any side effects to actually occur.
 
 #### Simulation
 
-Simulating `Cmd.none` always returns null.
+Simulating `Cmd.none` always returns an empty array.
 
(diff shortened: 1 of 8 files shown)


bdwain:
I'm confused why you  need to check the type of the result now. Simulating Cmd.run and Cmd.action never returns an array, and Cmd.list always does. I guess simulating map could vary, but maybe we could limit the breaking change to just that?

laszlopandy:
On second thought, it's not worth breaking anything here. My intention was to make the `simulate` APIs consistent so you don't have to remember what it returns. You just know it always returns an array of actions.
On the other hand, simulate is probably only used in the tests and probably only when you know exactly what command you are dealing with. I mean, `simulate` is probably not often used in a generic context.

PR #240, opened 2020-04-20
"npm run format" auto-fixes all prettier issues

laszlopandy:


bdwain:
what's the reasoning for this one?

laszlopandy:
@bdwain Ok I just found the `npm run lint -- --fix` option. Nevermind 🤦

bdwain:
got it. we can wrap `yarn lint --fix` as `yarn format` if you want. (I want to be consistent with yarn vs npm since running npm install would be problematic in a yarn package). 

I never added it before because i usually just run the linter in my IDE when I change a file.

Issue #243, opened 2020-04-23
Fix Loop generics for TypeScript


nweber-gh:
The exported typings for the Loop object do not appear to be correct; the current implementation requires that an action be provided as one of the generics, even though many loops do not cause an action to be dispatched.

from `redux-loop/index.d.ts` (version 5.1.1)
```
export type Loop<S, A extends Action> = [S, CmdType<A>];
```

Example:
```
type AppState = {};
type FooAction = Action<'fooStuff'>;

function fooSideEffect(isFoo: boolean): void {
  isFoo && console.log("We got that foo.");
}

function handleMyAction(state: AppState, action: FooAction): Loop<AppState> {
  return loop(
    { ...state, foo: 'yes' },
    Cmd.run(fooSideEffect, {
      args: [true]
    })
  );
}
```

This causes an error where the return type of the function is specified.
> Generic type 'Loop' requires 2 type argument(s).

Using this signature fixes the issue but it doesn't match the actual implementation of what's happening.

```
function handleMyAction(state: AppState, action: FooAction): Loop<AppState, Action> {
```

nweber-gh:
@laszlopandy Do you know if this is an actual issue, or am I misunderstanding the `Loop` type?

laszlopandy:
Yes, a `Cmd.run` which doesn't specify any actions creators and doesn't use `dispatch()` doesn't need to care about the action type.
We could consider making the second type parameter to `Loop` optional. The default value could be `{}` or `any`. What happens if you specify those instead of `Action`?

laszlopandy:
In fact `never` might work in this case as well.

nweber-gh:
Good call. A `never` seems to work, so perhaps we should default to that. I'll add that now and report back later today about whether or not that continues to work as expected.

nweber-gh:
So far this appears to be working as expected. We might want to consider `never` as a default in the various `Cmd` types as well. For example, `Cmd.run` also requires actions to be provided (for success and fail) even though they are optional.

Are there are commands, aside from `Cmd.action`, that *must* include an action to dispatch?

laszlopandy:
There are only `Cmd.action`, `Cmd.run` and `Cmd.none`, the others (ie. list, map) just wrap other commands.

[...]

PR #244, opened 2020-04-28
Adjust TypeScript definition for places where actions may not be dispatched

nweber-gh:
As discussed in issue https://github.com/redux-loop/redux-loop/issues/243 actions are not always dispatched from loops. This PR defaults the `Action` generic to `never` in the various spots where an action may not be dispatched.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -71,7 +71,6 @@ export interface RunCmd<SuccessAction extends Action = never, FailAction extends
 export type SequenceCmd<A extends Action = never> = ListCmd<A>;
 export type BatchCmd<A extends Action = never> = ListCmd<A>;
 
-
 export type CmdType<A extends Action = never> =
   | ActionCmd<A>
   | ListCmd<A>
@@ -103,7 +102,16 @@ export namespace Cmd {
   export function batch<A extends Action = never>(cmds: CmdType<A>[]): BatchCmd<A>;
   export function sequence<A extends Action = never>(cmds: CmdType<A>[]): SequenceCmd<A>;
 
-  export function list<A extends Action = never>(
+  export function list(
+    cmds: CmdType[],
+    options?: {
+      batch?: boolean;
+      sequence?: boolean;
+      testInvariants?: boolean;
+    }
+  ): ListCmd;
+
+  export function list<A extends Action>(
     cmds: CmdType<A>[],
     options?: {
       batch?: boolean;
@@ -112,7 +120,7 @@ export namespace Cmd {
     }
   ): ListCmd<A>;
 
-  export function map<A extends Action, B extends Action = never>(
+  export function map<A extends Action, B extends Action>(
     cmd: CmdType<B>,
     tagger: (subAction: B) => A,
     args?: any[]
@@ -129,16 +137,51 @@ export namespace Cmd {
   }
   export type PromiseResult<T> = T extends Promise<infer U> ? U : T;
 
+  export function run<Func extends (...args: any[]) => Promise<any> | any>(
+    f: Func,
+    options?: {
+      args?: ArgOrSymbol<Parameters<Func>>;
+      forceSync?: boolean;
+      testInvariants?: boolean;
+    },
+  ): RunCmd;
+
   export function run<
     Func extends (...args: any[]) => Promise<any> | any,
     SuccessAction extends Action = never,
+    >(
+    f: Func,
+    options?: {
+      args?: ArgOrSymbol<Parameters<Func>>;
+      successActionCreator: (value: PromiseResult<ReturnType<Func>>) => SuccessAction;
+      forceSync?: boolean;
+      testInvariants?: boolean;
+    },
+  ): RunCmd<SuccessAction, never>;
+
+  export function run<
+    Func extends (...args: any[]) => Promise<any> | any,
     FailAction extends Action = never,
     >(
     f: Func,
     options?: {
       args?: ArgOrSymbol<Parameters<Func>>;
-      failActionCreator?: (error: any) => FailAction;
-      successActionCreator?: (value: PromiseResult<ReturnType<Func>>) => SuccessAction;
+      failActionCreator: (error: any) => FailAction;
+      forceSync?: boolean;
+      testInvariants?: boolean;
+    },
+  ): RunCmd<never, FailAction>;
+
+  export function run<
+    Func extends (...args: any[]) => Promise<any> | any,
+    SuccessAction extends Action,
+    FailAction extends Action,
+    >(
+    f: Func,
+    options?: {
+      args?: ArgOrSymbol<Parameters<Func>>;
+      failActionCreator: (error: any) => FailAction;
+      successActionCreator: (value: PromiseResult<ReturnType<Func>>) => SuccessAction;
       forceSync?: boolean;
       testInvariants?: boolean;
     },
@@ -172,4 +215,5 @@ export function isLoop(test: any): boolean;
 
 export function getModel<S>(loop: S | Loop<S, AnyAction>): S;
 
-export function getCmd<A extends Action = never>(a: any): CmdType<A> | null;
+export function getCmd<A extends Action>(a: any): CmdType<A> | null;
+export function getCmd(a: any): CmdType | null;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,7 @@
 import { Action, ActionCreator, AnyAction, StoreEnhancer, Store } from 'redux';
 
+export type UnknownAction = AnyAction | never;
+
 export interface StoreCreator {
   <S, A extends Action>(
     reducer: LoopReducer<S, A>,
@@ -79,8 +81,6 @@ export interface RunCmd<
   readonly forceSync?: boolean;
   simulate(simulation: CmdSimulation): SuccessAction | FailAction;
 }
-
-type UnknownAction = AnyAction | never;
 
 export type CmdType =
   | ActionCmd<UnknownAction>
@@ -188,18 +188,18 @@ export namespace Cmd {
   ): RunCmd<SuccessAction, FailAction>;
 }
 
-export type ReducerMapObject<S, A extends Action = Action> = {
+export type ReducersMapObject<S, A extends Action = AnyAction> = {
   [K in keyof S]: LoopReducer<S[K], A>;
 };
 
 export function combineReducers<S, A extends Action = AnyAction>(
-  reducers: ReducerMapObject<S, A>
+  reducers: ReducersMapObject<S, A>
 ): LiftedLoopReducer<S, A>;
 
 export function mergeChildReducers<S>(
   parentResult: S | Loop<S>,
   action: AnyAction,
-  childMap: ReducerMapObject<S>
+  childMap: ReducersMapObject<S>
 ): Loop<S>;
 
 export function reduceReducers<S>(
@@ -212,13 +212,10 @@ export function reduceReducers<S, A extends Action>(
   ...reducers: Array<LoopReducerWithDefinedState<S, A>>
 ): LiftedLoopReducer<S, A>;
 
-export function liftState<S, A extends Action>(state: S | Loop<S>): Loop<S>;
+export function liftState<S>(state: S | Loop<S>): Loop<S>;
 
 export function isLoop(test: any): boolean;
 
 export function getModel<S>(loop: S | Loop<S>): S;
 
-export function getCmd<A extends Action, B extends Action = never>(
-  a: any
-): CmdType | null;
 export function getCmd(a: any): CmdType | null;
--- a/index.d.ts
+++ b/index.d.ts
@@ -130,9 +130,9 @@ export namespace Cmd {
       : T[K];
   };
 
-  export type PromiseResult<T> = T extends Promise<infer U> ? U : T;
-
   type RunFunc = (...args: any[]) => Promise<any> | any;
+
+  export type PromiseResult<T> = T extends Promise<infer U> ? U : T;
 
   export type RunOptions<
     Func extends RunFunc,
@@ -157,16 +157,13 @@ export namespace Cmd {
     >
   ): RunCmd<never, never>;
 
-  export function run<
-    Func extends (...args: any[]) => Promise<any> | any,
-    SuccessAction extends Action
-  >(
+  export function run<Func extends RunFunc, SuccessAction extends Action>(
     f: Func,
     options: Omit<RunOptions<Func, SuccessAction>, 'failActionCreator'>
   ): RunCmd<SuccessAction, never>;
 
   export function run<
-    Func extends (...args: any[]) => Promise<any> | any,
+    Func extends RunFunc,
     FailAction extends Action,
     FailReason = unknown
   >(
@@ -178,7 +175,7 @@ export namespace Cmd {
   ): RunCmd<never, FailAction>;
 
   export function run<
-    Func extends (...args: any[]) => Promise<any> | any,
+    Func extends RunFunc,
     SuccessAction extends Action,
     FailAction extends Action,
     FailReason = unknown
--- a/index.d.ts
+++ b/index.d.ts
@@ -188,14 +188,14 @@ export namespace Cmd {
 }
 
 export type ReducerMapObject<S, A extends Action = Action> = {
-  [K in keyof S]: LoopReducer<S[K]>;
+  [K in keyof S]: LoopReducer<S[K], A>;
 };
 
 export function combineReducers<S>(
   reducers: ReducerMapObject<S, any>
 ): LiftedLoopReducer<S, any>;
 
-export function combineReducers<S, A extends Action = AnyAction>(
+export function combineReducers<S, A extends Action>(
   reducers: ReducerMapObject<S, A>
 ): LiftedLoopReducer<S, A>;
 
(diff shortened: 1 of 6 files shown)


laszlopandy:
Thanks for working on this!
I haven't tried this out in my apps yet (but I plan to). For now I only have one comments after reading your diff.

I don't think we have to give a default of `never` value for `Cmd.run`. TypeScript allows you to give multiple type definitions for the same function. For example, the following tells TypeScript to first consider the first function (ie if there aren't any action handlers, use `RunCmd<never>`):
```typescript
  export function run(
    f: Function,
    options?: {
      args?: any[];
      forceSync?: boolean;
      testInvariants?: boolean;
    }
  ): RunCmd<never>;

  export function run<A extends Action, B extends Action>(
    f: Function,
    options?: {
      args?: any[];
      failActionCreator?: ActionCreator<A>;
      successActionCreator?: ActionCreator<B>;
      forceSync?: boolean;
      testInvariants?: boolean;
    }
  ): RunCmd<A>;
```

nweber-gh:
Good call on the overloads, that is a better idea.

laszlopandy:
Right "overloads" -- I knew there was a word for it that I was forgetting!

Anyway this was just my first attempt at crafting an overload. Perhaps we can improve it by making more which are more specific. I'm not sure how to determine what will work best; it's really just trial and error with these kind of overloads.

nweber-gh:
@laszlopandy I took a stab at the overloads. Let me know what you think! I agree that it'll take some iterating to get this right.

index.d.ts
```diff
@@ -8,18 +8,18 @@ export interface StoreCreator {
   ): Store<S>;
 }
 
-export type Loop<S, A extends Action> = [S, CmdType<A>];
+export type Loop<S, A extends Action = never> = [S, CmdType<A>];
 
-export interface LoopReducer<S, A extends Action = AnyAction> {
-  (state: S | undefined, action: A, ...args: any[]): S | Loop<S, A>;
+export interface LoopReducer<S, A extends Action = AnyAction, B extends Action = A> {
```
nweber-gh, line 13: @laszlopandy I added this so that you can optionally delineate between the set of actions that a reducer consumes vs the set of actions that will be dispatched via loops. I think that this is needed since the set of actions being dispatched may not match, or even be a subset of, the actions being consumed. The Loop actions default to the consumed actions though, so this remains backwards compatible. Thoughts?

bdwain: i think we should consider #247 when changing LoopReducer

index.d.ts
```diff
@@ -8,18 +8,22 @@ export interface StoreCreator {
   ): Store<S>;
 }
 
-export type Loop<S, A extends Action> = [S, CmdType<A>];
+export type Loop<S, A extends Action = never> = [S, CmdType<A>];
 
-export interface LoopReducer<S, A extends Action = AnyAction> {
+export interface LoopReducerDeprecated<S, A extends Action = AnyAction> {
```
bdwain, line 13: it's probably worth throwing in a comment explaining why. Maybe link to https://github.com/reduxjs/redux/issues/3580 and copy a buggy example from the issue as a comment?

```ts
type MyActionTypes = ActionType1 | ActionType2; //implicitly ActionType1 | ActionType2 | never

const MyReducer: Reducer<MyState, MyActionTypes> = (state = initState, action) => {
  switch(action.type){
    case actionTypes.TYPE_1:
      return handleType1(state, action); //infers that action is ActionType1

    case actionTypes.TYPE_2:
      return handleType2(state, action); //infers that action is ActionType2
    
	default:
	  return action; //infers that action type is never, so it does not complain that the return type is invalid because it assumes this will never happen anyway. you can also just skip the default case entirely and it won't complain
  } 
};
```

index.d.ts
```diff
@@ -8,18 +8,22 @@ export interface StoreCreator {
...
 
-export interface LoopReducerWithDefinedState<S, A extends Action = AnyAction> {
-  (state: S, action: A, ...args: any[]): S | Loop<S, A>;
+export interface LoopReducer<S, LoopActions extends Action = never> {
+  (state: S | undefined, action: AnyAction, ...args: any[]): S | Loop<S, LoopActions>;
 }
 
-export interface LiftedLoopReducer<S, A extends Action = AnyAction> {
-  (state: S | undefined, action: A, ...args: any[]): Loop<S, A>;
+export interface LoopReducerWithDefinedState<S, A extends Action = AnyAction, B extends Action = A> {
```
bdwain, line 21: should we make a deprecated copy of this one too?

index.d.ts
```diff
@@ -8,18 +8,22 @@ export interface StoreCreator {
...
+  (state: S, action: A, ...args: any[]): S | Loop<S, B>;
+}
+
+export interface LiftedLoopReducer<S, A extends Action = AnyAction, B extends Action = A> {
```
bdwain, line 25: same here

bdwain: oh actually this one should just be updated to the new pattern and we can rename this one deprecated

index.d.ts
bdwain: oh wait this wasnt updated. we prob want to update all of them and make deprecated copies

index.d.ts
```diff
@@ -68,11 +72,10 @@ export interface RunCmd<SuccessAction extends Action, FailAction extends Action
 }
 
 //deprecated types
-export type SequenceCmd<A extends Action> = ListCmd<A>;
-export type BatchCmd<A extends Action> = ListCmd<A>;
-
+export type SequenceCmd<A extends Action = never> = ListCmd<A>;
```
bdwain, line 75: batch and sequence (the cmds, not the options on Cmd.list) can actually just be deleted. they were removed in v5 but i guess we forgot the types.

index.d.ts
```diff
@@ -129,6 +141,41 @@ export namespace Cmd {
   }
   export type PromiseResult<T> = T extends Promise<infer U> ? U : T;
 
+  export function run<Func extends (...args: any[]) => Promise<any> | any>(
```
bdwain, line 144: doesn't `any` include `Promise<any>` anyway? why say `any | Promise<any>`?

nweber-gh:
Did a push with the default action enforcement discussed in https://github.com/redux-loop/redux-loop/issues/247

index.d.ts
nweber-gh: Not sure - I didn't modify that part since I don't know is there's a historical reason for is.

My guess is that the author was just being more explicit to be descriptive since, yes, `any` should include `Promise<any>`.

index.d.ts
```diff
@@ -8,18 +8,20 @@ export interface StoreCreator {
   ): Store<S>;
 }
 
-export type Loop<S, A extends Action> = [S, CmdType<A>];
+type WithDefaultActionHandling<T> = T | Action<'ENFORCE_DEFAULT_HANDLING'>;
```
bdwain, line 11: can you throw on a prefix like @@REDUX_LOOP/ENFORCE_DEFAULT_HANDLING to avoid name conflicts

bdwain: also would it make sense to say `T extends AnyAction`?

index.d.ts
```diff
@@ -129,6 +131,41 @@ export namespace Cmd {
...
+      testInvariants?: boolean;
+    },
+  ): RunCmd;
+
+  export function run<
+    Func extends (...args: any[]) => Promise<any> | any,
+    SuccessAction extends Action = never,
+    >(
+    f: Func,
+    options?: {
```
bdwain, line 148: these options are not optional because  the success action creator is required

index.d.ts
```diff
@@ -129,6 +131,41 @@ export namespace Cmd {
...
+    },
+  ): RunCmd<SuccessAction, never>;
+
+  export function run<
+    Func extends (...args: any[]) => Promise<any> | any,
+    FailAction extends Action = never,
+    >(
+    f: Func,
+    options?: {
+      args?: ArgOrSymbol<Parameters<Func>>;
```
bdwain, line 162: same here

index.d.ts
```diff
@@ -137,8 +174,8 @@ export namespace Cmd {
     f: Func,
     options?: {
```
bdwain, line 175: same here

index.d.ts
```diff
@@ -129,6 +131,41 @@ export namespace Cmd {
...
+    options?: {
+      args?: ArgOrSymbol<Parameters<Func>>;
+      forceSync?: boolean;
+      testInvariants?: boolean;
+    },
+  ): RunCmd;
+
+  export function run<
+    Func extends (...args: any[]) => Promise<any> | any,
+    SuccessAction extends Action = never,
```
bdwain, line 145: shouldn't there be no default here? Otherwise it's just the same as the first overload

index.d.ts
```diff
@@ -129,6 +131,41 @@ export namespace Cmd {
...
+      args?: ArgOrSymbol<Parameters<Func>>;
+      successActionCreator: (value: PromiseResult<ReturnType<Func>>) => SuccessAction;
+      forceSync?: boolean;
+      testInvariants?: boolean;
+    },
+  ): RunCmd<SuccessAction, never>;
+
+  export function run<
+    Func extends (...args: any[]) => Promise<any> | any,
+    FailAction extends Action = never,
```
bdwain, line 158: same here no default

index.d.ts
```diff
@@ -129,6 +131,41 @@ export namespace Cmd {
   }
   export type PromiseResult<T> = T extends Promise<infer U> ? U : T;
 
+  export function run<Func extends (...args: any[]) => Promise<any> | any>(
+    f: Func,
+    options?: {
```
bdwain, line 136: you could avoid repeating the options type by doing something like this.

```ts
type RunFuncType =(...args: any[]) => Promise<any> | any;
export type RunOptions<Func extends RunFuncType, SuccessAction, FailAction> = {
  args?: ArgOrSymbol<Parameters<Func>>;
  failActionCreator: (error: any) => FailAction;
  successActionCreator: (value: PromiseResult<ReturnType<Func>>) => SuccessAction;
  forceSync?: boolean;
  testInvariants?: boolean;
};

export function run<Func extends RunFuncType>(
  f: Func,
  options?: Omit<RunOptions<Func, never, never>, 'successActionCreator' | 'failActionCreator'>
): RunCmd;

export function run<
  Func extends RunFuncType,
  SuccessAction extends Action,
  >(
  f: Func,
  options: Omit<RunOptions<Func, SuccessAction, never>, 'successActionCreator'>
): RunCmd<SuccessAction, never>;

export function run<
  Func extends RunFuncType,
  FailAction extends Action,
  >(
  f: Func,
  options: Omit<RunOptions<Func, never, FailAction>, 'failActionCreator'>,
): RunCmd<never, FailAction>;

export function run<
  Func extends RunFuncType,
  SuccessAction extends Action,
  FailAction extends Action,
  >(
  f: Func,
  options: RunOptions<Func, SuccessAction, FailAction>,
): RunCmd<SuccessAction, FailAction>;
```

index.d.ts
```diff
@@ -100,8 +95,15 @@ export namespace Cmd {
   export type GetState = <S>() => S;
 
   export function action<A extends Action>(action: A): ActionCmd<A>;
-  export function batch<A extends Action>(cmds: CmdType<A>[]): BatchCmd<A>;
-  export function sequence<A extends Action>(cmds: CmdType<A>[]): SequenceCmd<A>;
+  
+  export function list(
+    cmds: CmdType[],
+    options?: {
```
bdwain, line 101: same here with ways  to consolidate list options

index.d.ts
nweber-gh: Yep, both good points.

index.d.ts
nweber-gh: Oh yeah, that's a good catch. It was a remnant from the original, non-overload approach I think. Ditto for the failure action default - the point of the overload is that if you're using it that you have one.

nweber-gh: While we're on the topic though... I actually don't know how / if TS could tell the difference between these two overloads.

```
  export function run<
    Func extends (...args: any[]) => Promise<any> | any,
    SuccessAction extends Action,
    >(
    f: Func,
    options?: {
      args?: ArgOrSymbol<Parameters<Func>>;
      successActionCreator: (value: PromiseResult<ReturnType<Func>>) => SuccessAction;
      forceSync?: boolean;
      testInvariants?: boolean;
    },
  ): RunCmd<SuccessAction, never>;

  export function run<
    Func extends (...args: any[]) => Promise<any> | any,
    FailAction extends Action,
    >(
    f: Func,
    options?: {
      args?: ArgOrSymbol<Parameters<Func>>;
      failActionCreator: (error: any) => FailAction;
      forceSync?: boolean;
      testInvariants?: boolean;
    },
  ): RunCmd<never, FailAction>;
```

They both take the same number of generics and the generics are the same type. I'm betting that this won't work.. We probably need to delete both of those and just have one signature that takes no actions and one that takes two actions.

```
  export function run<Func extends (...args: any[]) => Promise<any> | any>(
    f: Func,
    options?: {
      args?: ArgOrSymbol<Parameters<Func>>;
      forceSync?: boolean;
      testInvariants?: boolean;
    },
  ): RunCmd;

  export function run<
    Func extends (...args: any[]) => Promise<any> | any,
    SuccessAction extends Action,
    FailAction extends Action,
    >(
    f: Func,
    options?: {
      args?: ArgOrSymbol<Parameters<Func>>;
      failActionCreator: (error: any) => FailAction;
      successActionCreator: (value: PromiseResult<ReturnType<Func>>) => SuccessAction;
      forceSync?: boolean;
      testInvariants?: boolean;
    },
  ): RunCmd<SuccessAction, FailAction>;
```

Thoughts?

index.d.ts
nweber-gh: I did something similar. Lemme know what you think.

index.d.ts
nweber-gh: I pushed up what I'm thinking in commit https://github.com/redux-loop/redux-loop/pull/244/commits/c54b476487d23f7532fe9851c5eb0a47da1b304c.

bdwain: it can because one takes a failActionCreator and the other takes a success.

index.d.ts
```diff
@@ -1,5 +1,7 @@
 import { Action, ActionCreator, AnyAction, StoreEnhancer, Store } from 'redux';
 
+type Maybe<T> = T | undefined;
```
bdwain, line 3: Maybe is no longer used

index.d.ts
bdwain: @nweber-gh i would maybe just say `type ListOptions = {...}` so  you can reuse it in both overloads

index.d.ts
```diff
@@ -127,21 +131,54 @@ export namespace Cmd {
           ? typeof dispatch
           : T[K];
   }
+
   export type PromiseResult<T> = T extends Promise<infer U> ? U : T;
 
+  type RunFunc = (...args: any[]) => Promise<any> | any;
+
+  type RunOptions<
```
bdwain, line 139: should we just export this? maybe someone wants to type their options object.

index.d.ts
```diff
@@ -127,21 +131,54 @@ export namespace Cmd {
...
-      successActionCreator?: (value: PromiseResult<ReturnType<Func>>) => SuccessAction;
-      forceSync?: boolean;
-      testInvariants?: boolean;
-    },
+    options: Omit<RunOptions<Func, never, FailAction, FailReason>, 'successActionCreator'>,
+  ): RunCmd<never, FailAction>;
+
+  export function run<
+    Func extends (...args: any[]) => Promise<any> | any,
+    SuccessAction extends Action = never,
```
bdwain, line 176: SuccessAction and FailAction should not have defaults here. If you don't specify either it can be one of the other overloads

nweber-gh:
@bdwain @laszlopandy As I've been working through this, there's a fundamental thing that I'm wondering about.. Do we need to enforce type checking on actions emitted via Redux Loop? i.e., do we need to force the user to enumerate the types of actions that various commands can emit?

```
export type Loop<S, A extends Action = never> = [
  S,
  CmdType<A>
];

export interface LoopReducer<
  S,
  ReducerActions extends Action = AnyAction,
  LoopActions extends Action = never
> {
  (
    state: S | undefined,
    action: WithDefaultActionHandling<ReducerActions>,
    ...args: any[]
  ): S | Loop<S, LoopActions>;
}
```

Right now the PR has the slight alternation to master above, which indicates that a loop reducer returns a loop that is a set of known actions.  So one might have:

```
type ReducerActions = Action<'reducerAction1'>;
type LoopActions = Action<'loopAction1'>;

const myReducer: LoopReducer<FooState, ReducerActions, LoopActions> = (state, action) => {
  case "reducerAction1":
    return loop(state, Cmd.action('loopAction1'));
  default:
    return state;
}
```

And if you made a change..

```
const myReducer: LoopReducer<FooState, ReducerActions, LoopActions> = (state, action) => {
  case "reducerAction1":
    return loop(state, Cmd.list([Cmd.action('loopAction1'), Cmd.action('differentAction')]));
  default:
    return state;
}
```

You would end up with an error since `differentAction` is not in the `LoopActions` definition.

I'm not sure that we need to be enforcing this. It doesn't really matter what actions Loop emits. It could be any action in the system and I don't see why we need to force the user to enumerate the types of actions Loop cares about.

I *do* think we should enumerate the types of actions handled, since not handling an expected action is probably a bug.

I also think we need to specify action information where we have action creators arguments (such as `RunCmd`) so that we can be type safe in providing the right types of arguments.

But.. It feels overly verbose to force the user to enumerate the types of actions that Loop is emitting into the Redux system.

If we dropped this enforcement, we could dramatically simplify the type system. Especially since we have commands that take multiple actions which makes doing the `CmdType` union pretty difficult, and really complicates the types.

Thoughts?

bdwain:
I think that makes sense @nweber-gh

nweber-gh:
Did a push with the proposed changes. Tests are clean. I also added actual linting for the TypeScript stuff, since the linter was all confused.

PR #251, opened 2020-10-21
upgrade TS and enhance cmd.run types

nweber-gh:
This enhances the Cmd.Run types so that it will properly understand more complex types. The main use case is for *optional* `getState` / `dispatch` (e.g., `function doStuff(getState?: Cmd.GetState)`).

The types only work on TS 4, so I upgraded that as well.

The tests started failing, but incorrectly so. They complained about unused `@ts-expect-error` usages, which is was flat out wrong. I am guessing that the testing package isn't compatible with TS 4. But nevertheless, the package is useless anyway. All it does is verify that the package compiles, which we can easily do by simply running `tsc`.


nweber-gh:
@bdwain , we probably want to do a major version bump since this updates the TS version.

index.d.ts
```diff
@@ -66,15 +66,21 @@ export interface SetTimeoutCmd<A extends Action = never> {
   readonly nestedCmd: CmdType;
   readonly delayMs: number;
   readonly scheduledActionCreator?: (timerId: number) => A;
-  simulate(timerId: number, nestedSimulation?: CmdSimulation | MultiCmdSimulation): A[] | A | null
+  simulate(
```
nweber-gh, line 69: I did an eslint fix on the file and this stuff changed.

bdwain:
so this is a breaking change i guess. which is fine people can stay on v6. there's not a lot of other new stuff.

nweber-gh:
That was my thought, though it's also maybe weird to have just definitions be a breaking change. I think that's safer though. It could just be a minor and not a patch though. That should be fine, right?  Will the default semver pattern update minors?  I forget.

bdwain:
eh i dont think we should treat definitions differently than the rest of the code. it won't make a difference for js consumers but few breaking changes affect everyone anyway.

PR #254, opened 2020-11-18
Typescript fixes: export WithDefaultActionHandling, fix generics

laszlopandy:
1. Export `WithDefaultActionHandling`, otherwise it is impossible to explicitly specify a correct type for your reducer's arguments, like so:
   ```
   function reducer(state: State, action: WithDefaultActionHandling<Action>): State | Loop<State> { ... }
   ```
2. Allow specifying the generic types of `Dispatch` and `GetState`.

Both of these changes are fully backwards compatible.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -126,8 +126,8 @@ export namespace Cmd {
   export const dispatch: unique symbol;
   export const getState: unique symbol;
   export const none: NoneCmd;
-  export type Dispatch = <A extends Action>(a: A) => Promise<A>;
-  export type GetState = <S>() => S;
+  export type Dispatch<A extends Action = AnyAction> = (a: A) => Promise<A>;
+  export type GetState<S = any> = () => S;
 
   export function action<A extends Action>(action: A): ActionCmd<A>;
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -10,7 +10,7 @@ export interface StoreCreator {
   ): Store<S>;
 }
 
-type WithDefaultActionHandling<T extends AnyAction> =
+export type WithDefaultActionHandling<T extends AnyAction> =
   | T
   | Action<'@@REDUX_LOOP/ENFORCE_DEFAULT_HANDLING'>;
 


nweber-gh (approved):

bdwain:
I'm not sure I agree with this solution. WithDefaultActionhandling was meant to be used inside of the reducer type itself. you don't need to specify argument types if you do that.

```ts
const reducer: LoopReducer<MyState, ActionTypesToHandle> = (
  state = initialState,
  action
) => {
```

the types of state and action there can be inferred, so the WithDefaultActionhandling type is not needed. If you have to specify that explicitly as a consumer I think we're doing something wrong. It's an implementation detail of how we require you to handle all action types. Not something the consumer should be concerned with.

Also, I tested the other changes to dispatch and getState and they broke my app, so I'm not sure they can work as-is

laszlopandy:
For dispatch and getState, can you give an example so I can write a test for it?

bdwain:
I have a function with this signature

```ts
export function delayedDispatch<T extends unknown[]>(
  isTimeout: boolean,
  ms: number,
  dispatch: DispatchFn,
  actionCreator: (...params: T) => AnyAction,
  ...params: T
): number
```

and i get a type error here

```ts
  return loop(
    state,
    Cmd.run(delayedDispatch, {
      args: [true, 123, Cmd.dispatch, actionCreator, false],
    })
  );
```

there's a type error on action creator saying it's not assignable to unique symbol. i think these changes broke the way the ArgsOrSymbol helper works.

bdwain:
What exactly are you having trouble doing that you needed these changes?

laszlopandy:
I have a command helper which makes it easier to use `getState` and `dispatch` by passing both in an object I called `StoreProxy`:
```typescript
export interface StoreProxy<State, Action extends Redux.Action> {
    getState: Cmd.GetState<State>;
    dispatch: Cmd.Dispatch<Action>;
}
```

But these types are so simple, in this case it is easier to just inline them:
```typescript
export interface StoreProxy<State = unknown, Action extends Redux.Action = Redux.Action> {
    getState: () => State;
    dispatch: (a: Action) => void;
}
```

I didn't imagine this would break anything; I just thought it would be more flexible if the generic type could be specified. We can revert that change.

laszlopandy:
To avoid mixing two threads, I created an issue to discuss the `WithDefaultActionHandling` change: https://github.com/redux-loop/redux-loop/issues/255

PR #262, opened 2021-03-26
Add types overwrites for react-redux

DanielOrtel:


react/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "redux-loop-react",
```
bdwain, line 2: call this redux-loop/react. then you can import it like that

DanielOrtel: that's illegal in package.json, it can only have `/` if it's scoped

bdwain:
No it’s not illegal. Only if you’re publishing the package on its own.
We’re not publishing this separately.

DanielOrtel:
> No it’s not illegal. Only if you’re publishing the package on its own. We’re not publishing this separately

Alright, I can change it to that, but a simple yarn add will complain that `package.json: Name contains illegal characters`.

bdwain:
> > No it’s not illegal. Only if you’re publishing the package on its own. We’re not publishing this separately
> 
> Alright, I can change it to that, but a simple yarn add will complain that `package.json: Name contains illegal characters`.

that's fine. you wouldn't `yarn add redux-loop/react`. You would only add redux-loop. the `/react` import target is just an additional thing you can import if you have the redux loop dep. Our e2es are set up this way too.

react/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "redux-loop/react",
+  "description": "React specific declarations for redux-loop",
```
bdwain, line 3: we shouldn;t have all of these fields in this file. we really only need name and types. and probably set private to true. we don't want to be publishing this on its own.

react/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "redux-loop/react",
```
bdwain, line 2: on second thought, i kind of thing this package should be `redux-loop/react-redux` since that's the types we're overriding. And the folder should be react-redux as well.

react/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "description": "React specific declarations for redux-loop",
+  "homepage": "https://github.com/redux-loop/redux-loop",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/redux-loop/redux-loop.git"
+  },
+  "bugs": {
+    "url": "https://github.com/redux-loop/redux-loop/issues"
+  },
+  "files": [
```
bdwain, line 12: we should actually make sure to include the whole react-redux folder in the files array in the main package.json. not this one.

bdwain:
we should mention this in the docs. maybe create a typescript file in docs/tutorial (and reference it in the toc files) just explaining what this does?

react/package.json
nweber-gh: I was going to suggest the same.

DanielOrtel: yeah, that makes sense

react-redux/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "redux-loop/react-redux",
+  "private": true,
+  "types": "./index.d.ts",
+  "peerDependencies": {
```
bdwain, line 5: can you delete the peer deps here. they won't be evaluated anyway

docs/api-docs/TypeScript.md
```diff
@@ -0,0 +1,34 @@
+# `Typescript`
+
+## Usage with `react-redux`
+
+In order to use `redux-loop` within a  TypeScript React project, you'll need to import the type
```
nweber-gh, line 5: I was thinking that maybe they could directly pull from `node_modules` in the `tsconfig.json`. But maybe that would be bad practice - I suggested to try that where we implement this to see how it looks/feels. It feels lame to have to make a file just to have a single line import.. But that is *probably* better so that node figures out how to resolve stuff rather than us hardcoding the module's file system path in `tsconfig.json`. What do you guys think @DanielOrtel @bdwain ?

DanielOrtel: there are a lot of different package managers and with `pnp`, this wouldn't be possible. I think the docs should show a solution that works regardless of the package manager.

nweber-gh: Ah, that's a good point.

nweber-gh (approved):

bdwain (approved):

Issue #252, opened 2020-11-06
Suggestion for test utilities


abradley2:
When testing applications that use a nested Elm architecture, it's often difficult to test commands in the way prescribed by the docs here (I've changed up the example so it's only testing for the existence of a Cmd after the reducer is run, which is a common use-case we have):
```ts
test('when the loadingStart action fires, the fetchDetails RunCmd is sent', (t) => {
  const state = { loading: false };

  const fetchDetailsCmd = getCmd( reducer(state, loadingStart(1)) )

  t.deepEqual(
    fetchDetailsCmd,
    Cmd.run(fetchDetails, {
      successActionCreator: loadingSuccess,
      failActionCreator: loadingError,
      args: [1]
    })
  )
})
```

If I'm testing a high level reducer, that has a nested reducer for a page, which may have a nested reducer for a big form on that page, I'm often dealing with a `CmdType` returned by a reducer that's a `ListCmd` with a number of `MapCmd`'s with the actual `RunCmd` or `ActionCmd` that I'm looking for a number of levels down. 

So doing a `deepEqual` on the result of the `Cmd` returned from the reducer will fail if I add another effect, making it a `ListCmd` (if I don't go back and update my test assertion's deepEqual). And if I'm just testing for the _existence_ of some effect I want to happen in a big stack of `ListCmd`'s and `MapCmd`'s, it becomes an annoying fixture to mock out. I just want to check that I'm triggering some effect, but the tests needs to be concerned with any other sibling effects that may happen.

Being strict about a `Cmd` result can be useful, but often what I want is to just ensure _some effect happens_.

My solution has been helper functions that let me run down a `CmdType` through any `MapCmd`'s and `ListCmd`'s so I can test for the existence of a `RunCmd` or an `ActionCmd` without having to worry 

Here's what I've been using and it's been working quite well.
```ts
function findRunCmd (cmd: CmdType, func: Function): CmdType | undefined {
  if (cmd.type === 'LIST') {
    return cmd.cmds.reduce(
      (acc: CmdType | undefined, cur: CmdType) => acc || findCmd(cur, func),
       undefined
    )
  }
  if (cmd.type === 'MAP') {
    return findCmd(cmd.nestedCmd, func)
  }
  if (cmd.type === 'RUN') {
    return cmd.func === func ? cmd : undefined
  }
}
```

The test case in the docs then looks something like this:
```ts
test('when the loadingStart action fires, the fetchDetails RunCmd is sent', (t) => {  
[...]

PR #263, opened 2021-03-28
Add `flattenCmd`

abradley2:
[Motivated by the discussion here](https://github.com/redux-loop/redux-loop/issues/252), we have an application making extensive use of `redux-loop`'s `Cmd.map` and `Cmd.list` functionality to support sub-modules (very similar to how Elm apps are traditionally structured) and want to test commands in a convenient way.

[Example fork used is published here](https://www.npmjs.com/package/@abradley2/redux-loop). In our tests we had a dozen situations where we used a similar `getRunCmd` function that did much of the logic in `flattenCmd` and then did a `.find`

Implementing that function now looks something like:
```
import { flattenCmd, CmdType, RunCmd } from '@abradley2/redux-loop'

export const getRunCmd = (cmd, func) => flattenCmd(cmd).find((c) => c.type === 'RUN' && c.func === func)
```

### DIFF
--- a/docs/tutorial/Testing.md
+++ b/docs/tutorial/Testing.md
@@ -34,6 +34,49 @@ test('reducer works as expected', (t) => {
 > not equal within JavaScript, and so are best to avoid if you want to compare
 > effects in your tests.
 
+## Working with nested cmd objects
+
+You may need to test objects from the `Cmd` module that make use of
+`Cmd.list` or `Cmd.map`. Imagine in the above example the reducer is updated to
+perform multiple commands when loading starts. We want to test that the properly
+shaped `fetchDetails` cmd is called, but don't necessarily want to mock out every
+effect in `Cmd.list` for an assertion. `flattenCmd` will return all commands nested in
+`Cmd.list`, `Cmd.map`, `Cmd.setInterval` and `Cmd.timeout` in a coniently un-nested array that contains
+only `Cmd.none`, `Cmd.run`, and `Cmd.action` objects.
+
+```js
+import test from 'tape';
+import reducer, { fetchDetails } from './reducer';
+import { loadingStart, loadingSuccess, loadingError } from './actions';
+import { Cmd, loop, flattenCmd, getCmd } from 'redux-loop';
+
+test('reducer fetches data after loadingStart action', (t) => {
+  t.plan(2)
+  const state = { loading: false };
+
+  const resultCmd = getCmd(
+    reducer(state, loadingStart(1));
+  );
+
+  const fetchCmd = flattenCmd(resultCmd).find((cmd) => {
+    return cmd.successActionCreator === loadingSuccess
+  })
+
+  t.ok(fetchCmd)
+
+  t.deepEqual(
+    fetchCmd,
+    Cmd.run(fetchDetails, {
+      successActionCreator: loadingSuccess,
+      failActionCreator: loadingError,
+      args: [1]
+    })
+  )
+
+  t.end()
+});
+```
+
 ## Simulating cmd objects
 
 Occasionally you may find yourself in a situation where you need to pass more
@@ -71,3 +114,11 @@ You can simulate any cmd object to test the actions returned. Lists take
 arrays of simulations for their child cmds.
 
 [See detailed documentation about simulating Cmds](/docs/api-docs/cmds.md)
+
+## Testing nested cmd objects
+
+You may have cases where you are interested in 
+
+```
+
+```
--- a/docs/tutorial/Testing.md
+++ b/docs/tutorial/Testing.md
@@ -41,7 +41,7 @@ You may need to test objects from the `Cmd` module that make use of
 perform multiple commands when loading starts. We want to test that the properly
 shaped `fetchDetails` cmd is called, but don't necessarily want to mock out every
 effect in `Cmd.list` for an assertion. `flattenCmd` will return all commands nested in
-`Cmd.list`, `Cmd.map`, `Cmd.setInterval` and `Cmd.timeout` in a coniently un-nested array that contains
+`Cmd.list`, `Cmd.map`, `Cmd.setInterval` and `Cmd.timeout` in a conveniently un-nested array that contains
 only `Cmd.none`, `Cmd.run`, and `Cmd.action` objects.
 
 ```js
@@ -114,11 +114,3 @@ You can simulate any cmd object to test the actions returned. Lists take
 arrays of simulations for their child cmds.
 
 [See detailed documentation about simulating Cmds](/docs/api-docs/cmds.md)
-
-## Testing nested cmd objects
-
-You may have cases where you are interested in 
-
-```
-
-```
--- a/index.d.ts
+++ b/index.d.ts
@@ -266,3 +266,5 @@ export function isLoop(test: any): boolean;
 export function getModel<S>(loop: S | Loop<S>): S;
 
 export function getCmd(a: any): CmdType | null;
+
+export function flattenCmd(cmd: CmdType): CmdType[];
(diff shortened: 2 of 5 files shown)


src/cmd.js
```diff
@@ -434,6 +434,20 @@ const none = Object.freeze({
...
 
+export function flattenCmd(cmd) {
+  if (
+    cmd.type === cmdTypes.MAP ||
+    cmd.type === cmdTypes.SET_TIMEOUT ||
+    cmd.type === cmdTypes.SET_INTERVAL
+  ) {
+    return flattenCmd(cmd.nestedCmd);
+  }
+  if (cmd.type === cmdTypes.LIST) {
```
laszlopandy, line 445: `else if`?

src/cmd.js
```diff
@@ -434,6 +434,20 @@ const none = Object.freeze({
...
+    return cmd.cmds.flatMap(flattenCmd);
+  }
+  return [cmd];
+}
```
laszlopandy, line 449: Is there any use case to justify including `Cmd.none` values in the array? It might be more useful if the array consists only of run commands and action commands.

index.d.ts
```diff
@@ -266,3 +266,5 @@ export function isLoop(test: any): boolean;
 export function getModel<S>(loop: S | Loop<S>): S;
 
 export function getCmd(a: any): CmdType | null;
+
+export function flattenCmd(cmd: CmdType): CmdType[];
```
laszlopandy, line 270: The return type could be more specific: 
`ActionCmd<UnknownAction> | NoneCmd | RunCmd<UnknownAction, UnknownAction>`

src/cmd.js
abradley2: I can't really think of a use case other than a testing purpose which would still be covered by checking `expect(flatCmd).toEqual([])`. I've updated to remove `Cmd.none` values 👍

src/utils.js
```diff
@@ -4,6 +4,10 @@ export const flatten = (array) => {
   return concat.apply([], array);
 };
 
+export const flatMap = (array, mapFn) => {
+  return array.reduce((nextArray, cur) => nextArray.concat(mapFn(cur)), []);
```
abradley2, line 8: Since there are instructions in the docs on how to use this library with browsers that don't natively have `Promise` or `Symbol` I felt I should avoid using `Array.prototype.flatMap` as none of those browsers would support it either.

laszlopandy:
Thanks for the updates @abradley2!
@bdwain I think this would be a good addition. WDYT?

index.d.ts
```diff
@@ -266,3 +266,5 @@ export function isLoop(test: any): boolean;
 export function getModel<S>(loop: S | Loop<S>): S;
 
 export function getCmd(a: any): CmdType | null;
+
+export function flattenCmd(cmd: CmdType): RunCmd<UnknownAction, UnknownAction> | ActionCmd<UnknownAction>;
```
bdwain, line 270: you accidentally got rid of the array portion of the return type here.

src/cmd.js
bdwain: I'm torn on this. I kind of like keeping the original cmds in there. though i guess there's not much value. Why filter them out though? The general idea with a flattened list is to say it `contains` a command right? if someone wants to look for Cmd.none, we should not stop them

src/cmd.js
```diff
@@ -434,6 +434,21 @@ const none = Object.freeze({
   simulate: () => null,
 });
 
+export function flattenCmd(cmd) {
+  if (
+    cmd.type === cmdTypes.MAP ||
+    cmd.type === cmdTypes.SET_TIMEOUT ||
+    cmd.type === cmdTypes.SET_INTERVAL
+  ) {
+    return flattenCmd(cmd.nestedCmd);
```
bdwain, line 443: we lose some info here. A map cmd is more than just the nested Cmd. it's also the tagger function. Is there any way we can preserve that in this?

docs/tutorial/Testing.md
```diff
@@ -34,6 +34,49 @@ test('reducer works as expected', (t) => {
...
+only `Cmd.run` and `Cmd.action` objects.
+
+```js
+import test from 'tape';
+import reducer, { fetchDetails } from './reducer';
+import { loadingStart, loadingSuccess, loadingError } from './actions';
+import { Cmd, loop, flattenCmd, getCmd } from 'redux-loop';
+
+test('reducer fetches data after loadingStart action', (t) => {
+  t.plan(2)
```
bdwain, line 54: what is t.plan?

docs/tutorial/Testing.md
```diff
@@ -34,6 +34,49 @@ test('reducer works as expected', (t) => {
 > not equal within JavaScript, and so are best to avoid if you want to compare
 > effects in your tests.
 
+## Working with nested cmd objects
+
+You may need to test objects from the `Cmd` module that make use of
+`Cmd.list` or `Cmd.map`. Imagine in the above example the reducer is updated to
+perform multiple commands when loading starts. We want to test that the properly
+shaped `fetchDetails` cmd is called, but don't necessarily want to mock out every
```
bdwain, line 42: what do you mean mock out effects? You shouldn't need to mock any effect because they are not run

docs/tutorial/Testing.md
```diff
@@ -34,6 +34,49 @@ test('reducer works as expected', (t) => {
...
+
+test('reducer fetches data after loadingStart action', (t) => {
+  t.plan(2)
+  const state = { loading: false };
+
+  const resultCmd = getCmd(
+    reducer(state, loadingStart(1));
+  );
+
+  const fetchCmd = flattenCmd(resultCmd).find((cmd) => {
```
bdwain, line 61: it seems a little odd to search by successActionCreator only to then do a deep equal later. Why not just say

```ts
  expect(flattenCmd(resultCmd)).toContain(Cmd.run(fetchDetails, {
      successActionCreator: loadingSuccess,
      failActionCreator: loadingError,
      args: [1]
    }))
```

src/cmd.js
abradley2: For our tests we haven't ever really felt the need to consider the tagger, but I do see how that can be a relevant detail. I also suppose even `Cmd.timeout` and `Cmd.interval` could be.

[Here was the solution I landed on, I think it makes the function very agnostic to what level of detail people are seeking to test against](https://github.com/redux-loop/redux-loop/pull/263/files#diff-5a04a1234d0d58e4a4a0221acde6f8f8909110c0100ceefd6b75621c1e6e6a6cR67-R82)

docs/tutorial/Testing.md
abradley2: updated this language to be a bit more clear :+1:

docs/tutorial/Testing.md
abradley2: This and `t.plan` where `tape` testing library specific. I've updated the examples to use the "jest" assertions which make the example tests a bit more clear

docs/tutorial/Testing.md
```diff
@@ -34,6 +34,54 @@ test('reducer works as expected', (t) => {
...
 
+## Working with nested cmd objects
+
+You may need to test objects from the `Cmd` module that make use of
+`Cmd.list` or `Cmd.map`. Imagine in the above example the reducer is updated to
+perform multiple commands when loading starts. We want to test that the properly
+shaped `fetchDetails` cmd is included, but don't want to perform a `deepEqual` that 
+is sensitive to how we've nested that command. `flattenCmd` will return all commands nested in
+`Cmd.list`, `Cmd.map`, `Cmd.setInterval` and `Cmd.timeout` in a conveniently un-nested array that contains
+only `Cmd.run` and `Cmd.action` objects.
```
abradley2, line 45: oops, forgot to update this last line

PR #268, opened 2021-11-08
Add HandlerReturn type

zyvz:


nweber-gh (approved):

index.d.ts
```diff
@@ -266,3 +266,5 @@ export function isLoop(test: any): boolean;
 export function getModel<S>(loop: S | Loop<S>): S;
 
 export function getCmd(a: any): CmdType | null;
+
+export type HandlerReturn<S> = S | Loop<S>;
```
bdwain, line 270: should we call it ReducerHandlerReturn? Might make it more clear

zyvz: Its kinda wordy but I'm impartial

nweber-gh: The most idiomatic is probably `LoopReducerHandlerFnReturn`, though I agree with Baruch that we're getting 'wordy. `HandlerFnReturn`? `HandlerFunctionReturn`? I dunno. I'm mostly impartial. Maybe a comment above the type would help?

bdwain: Well i thought the name change was more helpful for the user of the library, not the author. Maybe let's leave it and we can change it if there's feedback

